PR #56, opened 2022-12-22
feat: add matviews for the other base tables

jitoquinto:
@emily-flambe adding matviews for the newer version of stackadapt sync


dbt-cta/stackadapt/models/2_partner_matviews/account_campaigns_stats.sql
```diff
@@ -0,0 +1,5 @@
+{{ config(
+    auto_refresh = false,
```
jitoquinto, line 2: @emily-flambe  did a little science, it looks like if you set `auto_refresh` to `false` then the mat-view refreshes when the dbt is ran 😮

michaelefisher: I think this is def the way we should do mat views, but can we confirm this:
> I THINK means that if we change the schema of the matview, we won't need to do all that silliness with dropping and rebuilding them.

Because that would be HUGE!

I mean, as we know, we're using mat views as a hack for simplifying access and storage (that is, storing data in partner projects). Their technical value is in refreshing automatically complex joins and stuff. Caching. We don't care about that, so it doesn't matter to me really where we refresh? And having one authoritative place seems superior in my book.

PR #57, opened 2022-12-22
fix: use 'ref' instead of 'source' to specify model dependencies for partner views

emily-flambe:
This is a small thing, but:

When running a sync for the first time, the partner dbt step would always fail first and then succeed, because it was waiting for the matviews to exist in order to create the views. This was always a bit mysterious, since the tables were right there in the `--depends_on:` bit at the top, but I THINK that only works with {{ ref('foo') }}, not with {{ source('bar','foo') }}. Seems arbitrary but what do I know.


jitoquinto:
oh wow this whole time i thought those `--depends_on` lines were just comments, didnt realize they actually set up dependencies!

jitoquinto (approved):

PR #61, opened 2023-01-11
fix: make sure facebook matviews dont break without full refresh macro

jitoquinto:
@emily-flambe This is the change i was thinking. Its the whole replace partitions with todays and yesterdays date with latest data from raw (todays data). What ya think? or should we just make the matviews `full-refresh = True`

### DIFF
--- a/dbt-cta/facebook_marketing/models/1_cta_base/account_report_base.sql
+++ b/dbt-cta/facebook_marketing/models/1_cta_base/account_report_base.sql
@@ -6,20 +6,17 @@
 {{ config(
     cluster_by = "_airbyte_emitted_at",
     partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
-    unique_key = '_ad_set_report_hashid'
+    partitions = partitions_to_replace,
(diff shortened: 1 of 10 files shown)


emily-flambe (approved):
This seems reasonable! We should check to make sure it does the behavior we expect, though (really need to start playing with `dbt test` or some strat for unit testing...) - or is it possible to verify that in Dev before deploying to Prod?

jitoquinto:
> This seems reasonable! We should check to make sure it does the behavior we expect, though (really need to start playing with `dbt test` or some strat for unit testing...) - or is it possible to verify that in Dev before deploying to Prod?

Yup [testing it out in dev](https://256d9e043bf64c2d9fda5c645b99c062-dot-us-central1.composer.googleusercontent.com/dags/facebook_marketing/grid)!  Its a bit different since it ran today with updated dbt so the matviews broke (underlying tables were replaced) but ill run it a few times to make sure its working as expected and were not seeing dupes

jitoquinto:
Alright tested it out in dev and looks to be running as expected. Bases tables are not getting recreated each run and partitions are being replaced

PR #64, opened 2023-01-19
feat: add action_builder dbt

jitoquinto:


emily-flambe (approved):
This is too many files for me to review all at once lol - but I approve! Actually I think this is "supposed" to be in Phase 2, but all the better that we have it done already!! should probably be clearer about what is in "Phase 1" vs "Phase 2" 🧐

jitoquinto:
> This is too many files for me to review all at once lol - but I approve! Actually I think this is "supposed" to be in Phase 2, but all the better that we have it done already!! should probably be clearer about what is in "Phase 1" vs "Phase 2" 🧐

Yeah im not entirely clear on what each Phase should do 🙃

PR #65, opened 2023-01-20
feat: init_dbt.sh should not create all those dang backup files

emily-flambe:
@michaelefisher I know there was some reason for adding the `-E` flag to these `sed` commands, but can you remind me what it was? It's causing the script to generate a bunch of backup files, which is something I was hoping we could avoid (see below).

If we need to keep it, I can add a step to remove files with the `-E` suffix, but it would be even cooler for them not to exist?

eg.:
```
spoke_sua_all_campaign_ab1.sql
spoke_sua_all_campaign_ab1.sql-E
spoke_sua_all_campaign_ab2.sql
spoke_sua_all_campaign_ab2.sql-E
spoke_sua_all_campaign_ab3.sql
spoke_sua_all_campaign_ab3.sql-E
...
```

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -39,7 +39,7 @@ PROJECT_ID=<CTA project id>
 PROJECT_NAME=<dev|prod>
 WORKSPACE_ID=<workspace ID>
 export WORKSPACE_ID="$WORKSPACE_ID";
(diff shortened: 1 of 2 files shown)


michaelefisher:
@emily-flambe wtf, it should definitely not be doing that. -E is for extended regex. Maybe it needs to be before the `-i`. But you can drop it I guess.

PR #68, opened 2023-01-25
feat: initialize action network dbt

jitoquinto:
https://techallies.atlassian.net/browse/DE-67

Here comes another big one 😓 basically just copies all the default normalization files from Airbyte, cleaned up a bit, and generated dbt files for partner materialization using the script i put on cacher.

I guess its still a bit unclear for me, for these sql mirror syncs is this the approach we are going with? will we be setting up direct db replication for other partners too and will the tables all match??


emily-flambe:
> I guess its still a bit unclear for me, for these sql mirror syncs is this the approach we are going with? will we be setting up direct db replication for other partners too and will the tables all match??

I think so? I think SQL Mirror strat will default to being a similar thing as other syncs (airbyte to BQ with our modified dbt, as in this PR). But we may need to re-evaluate if it turns out different partners have different tables and schemas...

dbt-cta/action_network/README.md
```diff
@@ -0,0 +1,3 @@
+# Action Network SQL Mirror
+
+TODO: Right up those docs
```
emily-flambe, line 3:
```suggestion
TODO: Write up those docs
```

emily-flambe (approved):

dbt-cta/action_network/README.md
jitoquinto: lmao thats embarrassing 😶‍🌫️

PR #70, opened 2023-02-07
fix: use correct dbt packages install path

emily-flambe:
@jitoquinto fixing dbt_project.yml in prod so that the dbt operators work (we had already applied this change in dev, but I neglected to submit the change to this repo to make it official).

### DIFF
--- a/dbt-cta/dbt_project.yml
+++ b/dbt-cta/dbt_project.yml
@@ -5,7 +5,7 @@ config-version: 2
 profile: 'default'
 
 model-paths: ["{{ env_var('SYNC_NAME') }}/models"]
-packages-install-path: "{{ env_var('SYNC_NAME') }}/dbt-packages"
+packages-install-path: "/tmp/{{ env_var('SYNC_NAME') }}/dbt-packages"
 
 models:
   default:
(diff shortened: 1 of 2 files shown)


dbt-cta/dbt_project.yml
```diff
@@ -5,7 +5,7 @@ config-version: 2
 profile: 'default'
 
 model-paths: ["{{ env_var('SYNC_NAME') }}/models"]
-packages-install-path: "{{ env_var('SYNC_NAME') }}/dbt-packages"
+packages-install-path: "/tmp/{{ env_var('SYNC_NAME') }}/dbt-packages"
```
jitoquinto, line 8: oh this was just me noodling around 😅 if were gonna try out the whole deps already in gcs this will have to be something different maybe even just `dbt-packages`?

emily-flambe: lol I've been inconsistent about this part of the discussion between this PR and the other one! Anyway, just using `"dbt-packages"` (no slash) seems to work! Updated the code in this PR accordingly 😎

emily-flambe:
@jitoquinto for final view to merge along with https://github.com/community-tech-alliance/airflow-dags/pull/139 !

jitoquinto (approved):

PR #79, opened 2023-02-14
Add POC dbt tests to mailchimp sync [DE-164]

jackwelty:
So far just `not_null` and `unique` tests on the ID col of the first mailchimp CTE!

### DIFF
--- a/dbt-cta/mailchimp/models/0_ctes/_0_ctes__models.yml
+++ b/dbt-cta/mailchimp/models/0_ctes/_0_ctes__models.yml
@@ -1,0 +1,12 @@
+version: 2
+
+models:
+
+  - name: campaigns_ab1
+
+    description: This model extracts fields from the raw json
+    columns:
+      - name: id
+        tests:
+          - unique
+          - not_null   


jackwelty:
`unique` checks on ID are passing for the models in `0_ctes` and `1_cte_base`, but want to verify that this is likely _not_ the case for other sources? (because we haven't deduped on the surrogate key at this point in the pipeline)

jitoquinto (commented):
This is really cool! Does this mean that dbt tests have to be done at the column level @jackwelty? i can see that getting out of hand with all the tables/columns we currently have on this repo 😓

jackwelty:
> This is really cool! Does this mean that dbt tests have to be done at the column level @jackwelty? i can see that getting out of hand with all the tables/columns we currently have on this repo 😓

Short answer is yes 😢. It depends on what we want to test, tests can be configured at both the table and the column level. A table level test would be like "Does this table have greater than X rows?", whereas most of the column level tests configured here are around uniqueness and not null-ness. There's a dbt code gen package that will help us write the schema files quicker, I can dig around to see if there is code to auto-generate tests.

emily-flambe:
>There's a dbt code gen package that will help us write the schema files quicker, I can dig around to see if there is code to auto-generate tests.

ooooo 👀 This tantalizes me greatly - I'll write up a ticket for you to look into dbt code generation strats @jackwelty (not necessarily to prioritize, but I don't want us to forget this ambition)

emily-flambe (approved):
merge away!

PR #87, opened 2023-02-23
feat: init Twilio dbt

jitoquinto:
Made some modifications to the Airbyte default normalization cause it did the whole break out into a ton of tables thing. 

Tested this out locally in Dev

- cta tables in this dataset - dev3869c056.twilio_cta_connector
- partner tables in this dataset - dev-partner-a-25693fa5.twilio


dbt-cta/twilio/models/0_ctes/available_phone_numbers_local_ab1.sql
```diff
@@ -0,0 +1,32 @@
...
+    {{ json_extract_scalar('_airbyte_data', ['lata'], ['lata']) }} as lata,
+    {{ json_extract_scalar('_airbyte_data', ['region'], ['region']) }} as region,
+    {{ json_extract_scalar('_airbyte_data', ['latitude'], ['latitude']) }} as latitude,
+    {{ json_extract_scalar('_airbyte_data', ['locality'], ['locality']) }} as locality,
+    {{ json_extract_scalar('_airbyte_data', ['longitude'], ['longitude']) }} as longitude,
+    {{ json_extract_scalar('_airbyte_data', ['iso_country'], ['iso_country']) }} as iso_country,
+    {{ json_extract_scalar('_airbyte_data', ['postal_code'], ['postal_code']) }} as postal_code,
+    {{ json_extract_scalar('_airbyte_data', ['rate_center'], ['rate_center']) }} as rate_center,
+    {{ json_extract('table_alias', '_airbyte_data', ['capabilities'], ['capabilities']) }} as capabilities,
+    {{ json_extract_scalar('_airbyte_data', ['capabilities','MMS'], ['capabilities','MMS']) }} as capabilities_MMS,
```
jitoquinto, line 20: oh oh it turns out unnesting nested json fields isnt too hard, just have to create a list with all the levels. So for this case the json looks like 
```json
{
  "capabilities":
      "MMS": true,
      "SMS": true,
      "voice": true
```
So if you want each individual field you gotta give the path to it as a list with the `json_extract_scalar` macro.

emily-flambe (approved):
Yeah wow there are a lot of tables here. I'm more or less trusting that your testing in dev is evidence enough that this is approvable 😎  And cutting down from 4 to 1 thread makes sense per our Slack convo that one time 🤔

PR #89, opened 2023-02-24
fix: make usvote base models incremental

jitoquinto:
Since the USVote base models were `1_cta_tables` they were getting dropped and rebuilt each run. This makes the matviews unhappy. So instead convert to incremental models with partition replacement.


huystuhh (approved):

emily-flambe:
@jitoquinto sorry I know this is already merged but I'm realizing I don't fully understand the following: If these models are still in 1_cta_tables, which materialize as `table` instead of `incremental` (per dbt_project.yml), then how does dbt now know that those models should be incremental? I see we added the partition replacement bits here and the `if is_incremental()` bit, but shouldn't `is_incremental()` resolve to false? Why does this fix work 🤔

jitoquinto:
> @jitoquinto sorry I know this is already merged but I'm realizing I don't fully understand the following: If these models are still in 1_cta_tables, which materialize as `table` instead of `incremental` (per dbt_project.yml), then how does dbt now know that those models should be incremental? I see we added the partition replacement bits here and the `if is_incremental()` bit, but shouldn't `is_incremental()` resolve to false? Why does this fix work 🤔

oh there no longer in `1_cta_tables` i changed the folder name to `1_cta_base`

PR #90, opened 2023-02-28
Refactor Snapchat Marketing materialized views, add tests [DE-206]

jackwelty:
This PR refactors our Snapchat Marketing materialized views, moving the dedupe logic to the `1_cta_base` models, so our materialized views can be simple `select *`s. I'm also adding tests as I go to ensure our dedupe logic is working as intended. Most of the models had airbyte-generated surrogate keys that were either incorrect or unnecessary, so I'm simplifying those as I go. I'll document the major changes to models below.

* Swapped primary key to `id`: A bunch of objects have just a straightforward `id` field, that per the [Airbyte Snapchat docs ](https://github.com/airbytehq/airbyte/blob/master/docs/integrations/sources/snapchat-marketing.md#supported-streams-and-sync-modes) is the primary key. For all these models, I've swapped `id` in as the `unique_key` in our incremental model config, and added relevant tests. List of models:
  - Adaccounts
  - Ads
  - Adsquads
  - Campaigns
  - Creatives
  - Media
  - Organizations
  - Segments

* `adsquads_targeting_base` / `adsquads_targeting` (mat view). This table was missing a relevant primary key. The auto-generated surrogate key is unique, but there's nothing remaining in the model to join this targeting info back to the ad squad that it applies to. I've gone ahead and added `ad_squad_id` to this series of models, and made that the primary key in the incremental materialization.
* `adsquads_targeting_geos_base` / `adsquads_targeting_geos`. Since I removed the `targeting_hashid`, I also had to tweak this table. We do still need a surrogate key here, but constructed slightly differently. We also weren't parsing any of the relevant geo info. I had to make some educated guesses about what we want the end state to be here. I think my structure will be robust to new geographic levels being added, and I've added some tests to check that my assumptions hold. One thing I'd like to add is a test to discover new geographic levels that might appear in our data without us knowing (the docs reference `electoral` as a geo-type, but it's not currently in our test data).

### TODO
- [ ] ~~Add a test to parse json keys in geo targeting to see if new geographies appear.~~ This seems like it could be very tricky, and we can rely on users requesting access to data at a different geo, if it exists.
- [x] Add "relationships" tests across a couple key models
- [ ] Add more docs

### Deployment thoughts
Since I'm changing the `unique_key` for the incremental strategy, I had to run all of these models (in `dev`) with the `--full-refresh` CLI flag. When we deploy this to prod, we'll need to full refresh all of our incremental tables, and likely drop and recreate our materialized views.

I think we'll also want to do some sort of end-to-end test on prod data (pulling the raw data into dev first, maybe?) to make sure there's nothing in another partner's data that is going to trip us up.


dbt-cta/snapchat_marketing/models/0_ctes/adsquads_targeting_ab3.sql
```diff
@@ -6,10 +6,7 @@
 -- depends_on: {{ ref('adsquads_targeting_ab2') }}
 select
     {{ dbt_utils.surrogate_key([
```
jackwelty, line 8: @emily-flambe I'm not using this surrogate key any more (the primary key of the table should be `ad_squad_id`). I kept this bit in for continuity's sake, but would it it be simpler if I just dropped the SK and the model entirely?

emily-flambe: ah! if we aren't using it then no need to keep the _ab3 step - just need to make sure any downstream reference(s) to `_ab3` are replaced with `_ab2`. In general if we can use a UPK other than the hashid that Airbyte generates in `_ab3`, we should preferentially use that instead and yeet `_ab3` - we just haven't taken the time to be super rigorous about that 🫠🫠🫠🫠

jackwelty: roger that! I will do that for all of these objects then, a bunch of these "simpler" models just have a straight `id` field we can use. Will increase the scope of this refactor a bit but I think it's worth it.

jackwelty:
Ok I stil have a couple of smaller TODOs left here, but I've got a [green run in airflow](https://256d9e043bf64c2d9fda5c645b99c062-dot-us-central1.composer.googleusercontent.com/dags/TEMP_snapchat_marketing/grid?dag_run_id=scheduled__2023-02-27T10%3A30%3A00%2B00%3A00&task_id=snapchat_marketing_partner_a.partner_a_transforms.0_partner_a_dbt_test_sources_cta) and I think the core logic of this (admittedly too big) PR is ready for review.

jitoquinto (commented):
lgtm! one question tho, are the models in `1_cta_base` working as expected? That folder has the incremental_strategy set to `insert_overwrite` so wondering if that might cause it to not use the unique key to merge?

jitoquinto:
> does this mean we should now remove the `is_incremental()` clauses? If I'm understanding the logic correctly, the current implementation will narrow our merge to rows that have come in in the last day... is there a chance that a given ad could appear in the sync on Day 1, not appear in Day 2, and then when it shows back up again in Day 3, we'll create a duplicate? Or is this not relevant since we've got the Airbyte syncs on `Full Refresh | Overwrite`?

@jackwelty yeah removing the incremental clauses sounds right to me too! We should probably just remove both the `is_incremental()` blocks and the `partitions_to_replace` up top too since we wont be replacing partitions. I dont think it'll run into any duplicates since merge should just _replace_ a record if its the same ad right? at least hoping thats what the unique_key means haha. 

Also, I've been writing my understanding of all these strategies [in this doc](https://docs.google.com/document/d/1qnXunxb9jR478_8beWKKoiciLkxy74VDZwJft8t1vWE/edit#heading=h.4c4f0mq7bj7e) if its helpful

jackwelty:
Ok changes made @jitoquinto! Tested and ✅ in dev Composer.

jitoquinto (approved):
lgtm!

kanelouise (approved):
Very fancy! Also learned a lot from the discussion here between you (@jackwelty) and @jitoquinto.

PR #94, opened 2023-03-06
fix (FB marketing): aggregate `views` metrics before left joining to `ads_insights_overall`

emily-flambe:
[DE-202](https://techallies.atlassian.net/browse/DE-202)

~I want to take a closer look at the ephemeral tables in the FB Marketing model. Won't do this in prod. (It wouldn't matter if we did, but I still don't want to pollute the prod CTA datasets with all the ephemeral tables...)~

Science has concluded and I think this is the fix that will make our dreams come true @jackwelty 👀 

[DE-202]: https://techallies.atlassian.net/browse/DE-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackwelty (changes requested):
Thanks for digging in here! I think the gist of the fix looks right to me. Some of the test queries from my investigation are appropriately returning a single row.

One note: do we want to sum() and aggregate the other metrics as well, besides `value` (`_1d_view`, `_7d_click`, etc.)?

emily-flambe:
> Thanks for digging in here! I think the gist of the fix looks right to me. Some of the test queries from my investigation are appropriately returning a single row.
> 
> One note: do we want to sum() and aggregate the other metrics as well, besides `value` (`_1d_view`, `_7d_click`, etc.)?

Good Q! Given that the data we are already delivering only use `value` I'm comfortable ignoring all those fields for now. We can always come back to those if a partner requests them in the future. 🐸

PR #96, opened 2023-03-13
feat: (Twilio) pre-aggregate in 0_ctes to avoid dupes in CTA base tables

emily-flambe:
As I was implementing dbt tests for the Twilio sync, I encountered an issue where sometimes the Airbyte incremental sync doesn't _quite_ work perfectly, and the same hashid ends up being synced multiple times. If we add a couple of CTEs to remove rows that were already emitted, then they won't be appended to the CTA base table, since the 1_cta_incremental models only append rows with new `_airbyte_emitted_at` values.

I've tested this by running dbt locally and copying the contents of the `_airbyte_raw` tables into `prod8b61f23e.TEMP_twilio_state_voices.` and it seems to be passing all the tests 👀  About to test this in dev composer :cat_typing:

@jackwelty @jitoquinto @kanelouise  this seems like a step towards our ambition of making sure the CTA base tables have all the deduping logic built in - curious whether you think this approach makes sense!


jackwelty (commented):
The deduping logic makes sense to me as implemented! My gut instinct would be to take the MAX() emitted at timbe, but I don't have experience with this particular source.

These are non-blocking, I just have a couple of clarifying questions about why this sync is different. This differs from my implementation of the Snapchat sync since none of those airbyte syncs are configured w/ `append`:
* A more meta question about airbtye: why is the alerts table configured as `Full Refresh | Append`?
* Why is `alerts_base` in `1_cta_base` as opposed to `1_cta_incremental`?
* Does the unique ID need to be changed in the other `base` models to the actual PK?

emily-flambe:
> * A more meta question about airbtye: why is the alerts table configured as `Full Refresh | Append`?
> * Why is `alerts_base` in `1_cta_base` as opposed to `1_cta_incremental`?

Oh! `alerts` was originally "Incremental Append," but Airbyte was delivering dupes in the raw data, which is why I switched it to "Full Refresh Overwrite" (there are only three rows anyway) - which is _also_ why the base model got moved to `1_cta_base`. But actually now I think maybe I should change it back and add the same deduping CTEs I added for the other `1_cta_incremental` models. Will make that change in a bit!

> * Does the unique ID need to be changed in the other `base` models to the actual PK?

You mean in the dbt? I'm actually not really sure - when I was trying to get the incremental refresh for `alerts` to work correctly, I tried setting `sid` as the PK (`unique_key` - that's what you're referencing right?), but it was still adding rows with duplicated `sid` values, so I gave up on trying to use that config. I don't think I have a satisfying answer for you here, beyond "maybe yes but also I don't really trust that setting a `unique_key` in dbt does what we want it to do, unless I'm doing it wrong and bad, which is definitely possible"

jitoquinto:
@emily-flambe is this because of the error thats like "Multiple matching entries for a unique key" or something like that?? (if it isnt because of this error just ignore the rest of this lol) I remember running into that once. I wanna say that means that the `_raw` table has duplicates cause the connectors incremental logic is a bit cribbed. Im wondering if we just add an `is_incremental()` block to our ctes so that it only grabs data with the latest `_airbyte_emitted_at`  we could avoid that error? and then the merging on unique key should take care of any de-duping?

dbt-cta/twilio/models/0_ctes/calls_ab4.sql
```diff
@@ -0,0 +1,16 @@
...
+    cluster_by = "_airbyte_emitted_at",
+    partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
+    unique_key = '_airbyte_ab_id',
+    tags = [ "top-level-intermediate" ]
+) }}
+-- SQL model to get latest updated dates for each hashid
+-- depends_on: {{ ref('calls_ab3') }}
+select
+    _airbyte_calls_hashid,
+    min(_airbyte_emitted_at) as _airbyte_emitted_at
```
jitoquinto, line 11: should we do max instead of min? wondering if its ever the case where the latest matching hash_id has any new data?

jitoquinto (approved):
Left some comments and questions but the de-duping logic makes sense!

dbt-cta/twilio/models/0_ctes/calls_ab4.sql
emily-flambe: Oh yeah, that's definitely something to consider. I'm not really sure the best way to approach this - the logic in the `1_cta_incremental` models seems to be: "add any new rows with more-recent `_airbyte_emitted_at` values". (That's what happens when we partition using _airbyte_emitted_at and set `{{ incremental_clause('_airbyte_emitted_at') }}`, right? 

But probably it makes more sense to have those use `date_updated` (or whatever it's called), huh? And all incremental streams should have some field like that, which we should probably use instead of `_airbyte_emitted_at`.

cc @jackwelty since I know your brain is also in this castle these days

jitoquinto (approved):
:shipit: :shipit: :shipit:

PR #97, opened 2023-03-14
Add tests to Facebook Marketing sync [DE-205]

jackwelty:
This PR adds tests to our Facebook Marketing sync, and refactors the sync slightly to move our dedupe logic upstream into the CTA models.

One call out: since this is a straight refactor, I did my best to maintain 100% column fidelity. This includes adding a copy of the `video_15_sec_watched_actions` column as `thruplays`, since it was implemented in the old `2_` models. I don't have context for why this column was added, but didn't want to remove it.

The companion PR that enables these tests is [here](https://github.com/community-tech-alliance/airflow-dags/pull/210).

## Deployment Notes
We'll have to refresh the materialized views, so deploying this to prod will result in some momentary downtime for partners.


emily-flambe (commented):
@jackwelty for the incremental models, are we confident that the current setup is delivering unique rows? Or do we need any pre-aggregation in CTEs to get "most recently updated" rows like we are implementing for Twilio ([example: alerts_ab3](https://github.com/community-tech-alliance/dbt-cta/blob/ec/dbt-test-twilio/dbt-cta/twilio/models/0_ctes/alerts_ab3.sql))?

dbt-cta/facebook_marketing/models/2_partner_matviews/_2_partner_matviews__models.yml
```diff
@@ -0,0 +1,806 @@
...
+      - name: impressions
+        description: ''
+      - name: spend
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+    tests:
+      - dbt_utils.equal_rowcount:
```
emily-flambe, line 31: oh fascinating!! this is my first time noticing this test - I wonder what it would mean if this one failed, though? It's basically just confirming that the matview refreshed and (at least in terms of rowcount) matches the table created by the `*_base` model? (Are there any other `compare_model` tests that might make sense?)

dbt-cta/facebook_marketing/models/3_partner_views/_3_partner_views__models.yml
```diff
@@ -0,0 +1,410 @@
...
+      - name: shares
+        description: ''
+      - name: landing_page_views
+        description: ''
+      - name: conversion_values
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+    tests:
+      - dbt_utils.unique_combination_of_columns:
```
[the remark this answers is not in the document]
jackwelty: yeah! `combination_of_columns` is kinda the same as creating and then ensuring the uniqueness of a surrogate key. I don't think it should replace us creaking SKs, but it's a nice double check.

dbt-cta/facebook_marketing/models/2_partner_matviews/_2_partner_matviews__models.yml
emily-flambe: Ahh I mean I agree with your logic here, this seems a fine starting point! Can always revise in the future if we get Older and Wiser 👍

jackwelty:
> @jackwelty for the incremental models, are we confident that the current setup is delivering unique rows? Or do we need any pre-aggregation in CTEs to get "most recently updated" rows like we are implementing for Twilio ([example: alerts_ab3](https://github.com/community-tech-alliance/dbt-cta/blob/ec/dbt-test-twilio/dbt-cta/twilio/models/0_ctes/alerts_ab3.sql))?


All of these airbyte syncs are set to `Full Refresh | Overwrite` so I think we _should_ be good (and the tests will fail quickly if we aren't!)

emily-flambe:
> > @jackwelty for the incremental models, are we confident that the current setup is delivering unique rows? Or do we need any pre-aggregation in CTEs to get "most recently updated" rows like we are implementing for Twilio ([example: alerts_ab3](https://github.com/community-tech-alliance/dbt-cta/blob/ec/dbt-test-twilio/dbt-cta/twilio/models/0_ctes/alerts_ab3.sql))?
> 
> All of these airbyte syncs are set to `Full Refresh | Overwrite` so I think we _should_ be good (and the tests will fail quickly if we aren't!). At least in the `dev` data, I was not observing any duplicate PKs for a given merge.

Oh!! Since that's the case, I think the base models should all be in 1_cta_base, and all use the same date-partition-replacement strat we use for full-refresh-overwrite models elsewhere - does that seem right to you? (Someday we ought to rename `1_cta_base` to something with more obvious semantics, like `1_cta_full_refresh` - actually be right back I'm gonna make a PR to add that model type and we can start switching over to it)

emily-flambe:
> > > > @jackwelty for the incremental models, are we confident that the current setup is delivering unique rows? Or do we need any pre-aggregation in CTEs to get "most recently updated" rows like we are implementing for Twilio ([example: alerts_ab3](https://github.com/community-tech-alliance/dbt-cta/blob/ec/dbt-test-twilio/dbt-cta/twilio/models/0_ctes/alerts_ab3.sql))?
> > > 
> > > 
> > > All of these airbyte syncs are set to `Full Refresh | Overwrite` so I think we _should_ be good (and the tests will fail quickly if we aren't!). At least in the `dev` data, I was not observing any duplicate PKs for a given merge.
> > 
> > 
> > Oh!! Since that's the case, I think the base models should all be in 1_cta_base, and all use the same date-partition-replacement strat we use for full-refresh-overwrite models elsewhere - does that seem right to you? (Someday we ought to rename `1_cta_base` to something with more obvious semantics, like `1_cta_full_refresh` - actually be right back I'm gonna make a PR to add that model type and we can start switching over to it)
> 
> I modeled this PR after what I did for [snapchat](https://github.com/community-tech-alliance/dbt-cta/pull/90). Switching to `_incremental` with the merge strategy and a primary key we feel good about means the deduping will happen at the `_incremental` step, always replacing old rows with newly inserted rows, which alleviates the need for any downstream deduping. This sync was originally using `1_cta_base` and we had to do the suboptimal groupby downsteam, which we're trying to replace.

Oh OK! Does this recap reflect your understanding or am I still off somewhere 🤔

If the table has a UPK and a "most recently updated" field: use CTEs to dedupe, and base models live in `1_cta_incremental`, which merges new data using the UPK

If the table has a UPK but we don't need to worry about only including "most recently updated" rows for each UPK - use `1_cta_incremental`, no need for CTE strats

If the table has no UPK, or at least we haven't definitively specified one - use Full Refresh Overwrite in Airbyte and create base models in `1_cta_full_refresh`

jackwelty:
> 

that is where my understanding is at atm, but I'm sure we'll learn more as we refactor more of these! I think I'd add to clarify:

> If the table has a UPK, a "most recently updated" field, **and Airbyte is emitting multiple rows per UPK**: use CTEs to dedupe, and base models live in 1_cta_incremental, which merges new data using the UPK
>
> If the table has a UPK we don't need to worry about only including "most recently updated" rows for each UPK (i.e.,  **Airbyte is emitting one row UPK**)- use 1_cta_incremental, no need for CTE strats

emily-flambe:
> > 
> 
> @emily-flambe that is where my understanding is at atm, but I'm sure we'll learn more as we refactor more of these! I think I'd add to clarify:
> 
> > If the table has a UPK, a "most recently updated" field, **and Airbyte is emitting multiple rows per UPK**: use CTEs to dedupe, and base models live in 1_cta_incremental, which merges new data using the UPK
> > If the table has a UPK we don't need to worry about only including "most recently updated" rows for each UPK (i.e.,  **Airbyte is emitting one row UPK**)- use 1_cta_incremental, no need for CTE strats

Awesome!! @jitoquinto we have a doc going somewhere that spells out stuff like this right? Can you drop a link to that and I'll add these bits in 🐸  meanwhile @jackwelty I shall approve this PR!

emily-flambe (approved):

emily-flambe:
@jackwelty I had to comment out a few `relationships:` tests that were failing - think it's worth creating a bug ticket to look into these? https://github.com/community-tech-alliance/dbt-cta/pull/97/commits/c307e405203c0f89a8a40dd1e7dfbde591a48eac

emily-flambe:
Ran these all locally to update VFP, SUA, and NDWA in prod! @jackwelty I made a couple of changes before merging this in: https://github.com/community-tech-alliance/dbt-cta/commit/c307e405203c0f89a8a40dd1e7dfbde591a48eac (commenting out some `relationships:` tests) and [f8a7922](https://github.com/community-tech-alliance/dbt-cta/pull/97/commits/f8a7922118e2528bd0e40fd112b6462d27b23a0f) to dedupe by hashid in ads_insights_overall. Please flag if any of those seem worth revisiting 👀

PR #98, opened 2023-03-22
feat: add dbt model `1_cta_full_refresh`

emily-flambe:
This is functionally equivalent to `1_cta_base` but the name makes it much more obvious what the model is for, which should resolve some confusion as we continue to refactor and develop our dbt projects!

Once this is merged in I'd like us to change `1_cta_base` over to `1_cta_full_refresh` and ultimately delete `1_cta_base` entirely - it seems possible that a straight find-and-replace on the folder names would do the trick, and I don't _think_ that would break things, but @jitoquinto @jackwelty @kanelouise would love for you to double-check my thinking there 🧐


jitoquinto (approved):
I like it! I thought i had a script to rename directories at some point but i cant find it, but it shouldn't be too hard to do that! theres a [python os function](https://www.tutorialspoint.com/python/os_rename.htm) that we can use to do it pretty easily

PR #99, opened 2023-03-23
feat: rename 1_cta_base => 1_cta_full_refresh

emily-flambe:
I'm gonna download a gzip of the current state of dbt-cta in dev before making these updates, since it spooks me a bit!


emily-flambe:
OK this seems to have gone over smoothly in dev! Definitely want your blessing @jitoquinto before merging this into prod - our deployment workflow doesn't remove deleted files, so I need to do a little manual stuff to make the files in GCS match our desired reality.

jitoquinto (approved):
:chonk:

PR #102, opened 2023-03-24
hotfix (USVoteFoundation): matviews should query from `_base` tables

emily-flambe:


emily-flambe:
YOLO-merging this to resolve a bug impacting the sync running in prod

PR #103, opened 2023-03-29
feat: add models for new ActBlue Stream

jitoquinto:
https://techallies.atlassian.net/browse/DE-238


dbt-cta/actblue/models/1_cta_full_refresh/cancelled_recurring_contributions_stream.sql
```diff
@@ -0,0 +1,42 @@
+{{ config(
```
emily-flambe, line 1: in the 1_cta_full_refresh models (formerly known as 1_cta_base), I noticed that here and in the other ActBlue models we don't we do the `set partitions_to_replace` date-partition-replacement strat - is that intentional? What are the circumstances under which we do that partition replacement strat? It seems like we inconsistently use the `set partitions_to_replace` bit and I can't recall if there's a specific reason for that 🤔 cc @kanelouise @jackwelty

jitoquinto: @emily-flambe alright i renamed the folder and everything is still working as expected! Also just gonna write out the stuff that was in slack here too. The `set_partitions_to_replace` is needed if the `incremental_strategy` is **insert_overwrite**. These models are overwriting the folder level configs in the `config` section and setting the `incremental_strategy` to **merge**

emily-flambe (approved):
Yay!! ty for moving the models, this calms my neurotic heart 😎

PR #107, opened 2023-04-03
Deploy Elementary via Composer

jackwelty:


dbt-cta/macros/cross_db_utils/datatypes.sql
```diff
@@ -75,9 +75,10 @@
 
 
 {# int  ------------------------------------------------- #}
-{% macro default__type_int() %}
+{# BQ does not have a "signed" data type, so this is commented out #}
+{# {% macro default__type_int() %}
     signed
```
emily-flambe, line 80: out of curiosity, was this causing a problem while it was not commented-out?

emily-flambe (approved):
I don't understand a lot of these configs on a deep spiritual level, but if they've been working and if they're necessary for elementary strats then :shipit:

dbt-cta/macros/cross_db_utils/datatypes.sql
jackwelty: It wasn't invoked anywhere in our project, but somewhere in an internal elementary thing, they called `CAST as default__type_int()`. This template rendered as `CAST as SIGNED` which failed. With this commented out, it renders as `CAST as INTEGER`, which is correct.

emily-flambe: oh that's sneaky!! thank you for the lore 🐸

PR #108, opened 2023-04-03
FIX:  ad account ids are not a fixed length [DE-255]

jackwelty:
The id field in the ad_account_base table starts with act_, so we previously used `right(id, 16)` to strip just the numbers but… apparently ad account IDs can be of varying length!

This PR replaces the `right()` function and `SPLIT`s on the underscore.


emily-flambe (approved):
excellent TY for the science, obviously did not realize this was a thing lol

kanelouise (approved):

PR #112, opened 2023-04-06
feat: change names of 1 and 2 folders

kanelouise:
The first step of [DE-223](https://techallies.atlassian.net/browse/DE-223) is to rename the folders to reflect our updated desired names:

`1_cta_full_refresh` or `1_cta_incremental` 
There should be no `1_cta_base`

`2_partner_matviews` 
The only reason `2_partner_views` should exist in a project is if `2_partner_matviews also exists.

If these things look true, then I'd love to get this merged into main before digging into each of the syncs to tidy up their dbt.

[DE-223]: https://techallies.atlassian.net/browse/DE-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kanelouise:
This PR is too confusing with merges, so I opened another PR doing the same thing on branch `kane/DE-223_rename_folders`

PR #113, opened 2023-04-06
feat: rename folders with new convention

kanelouise:
The first step of [DE-223](https://techallies.atlassian.net/browse/DE-223) is to rename the folders to reflect our updated desired names:

1_cta_full_refresh or 1_cta_incremental
There should be no 1_cta_base

2_partner_matviews
The only reason 2_partner_views should exist in a project is if `2_partner_matviews also exists.

If these things look true, then I'd love to get this merged into main before digging into each of the syncs to tidy up their dbt.


kanelouise:
@emily-flambe especially interested in your eyes on this one! Once we get this one merged in then the sync specific dbt work will begin for DE-223.

emily-flambe:
@kanelouise wowowow so many renamed files!! This seems correct to me - can we try a dev deployment first and re-run the dev DAGs to check for any fun surprises? If that goes well then I will approve 🙌

kanelouise:
The Mobilize sync worked. 

Action Builder and Action Network are going to take a bit more work (read: will have to be built first in Airbyte + added to the dev dag to test).

PR #114, opened 2023-04-07
hotfix: remove schema from configs in USVote Foundation sync

emily-flambe:
These should be specified in the config YAML read by daggy mcdagface, not specified within dbt. (The sync still delivers data successfully, but we want to use the same strat for all our syncs.)

cc @kanelouise - I think these get added in the Airbyte-generated dbt... I wonder if there's a way we can modify the init_dbt.sh script to remove the `schema = ...` lines? If not, then we will need to document somewhere that the person initializing the dbt for a new sync needs to make sure to remove these bits, since it's a very non-obvious step 🤔


kanelouise:
If you'd like to add it to [this ticket](https://techallies.atlassian.net/browse/DE-252), that's fine. Let's up the points, though 🙃

emily-flambe:
> If you'd like to add it to [this ticket](https://techallies.atlassian.net/browse/DE-252), that's fine. Let's up the points, though 🙃

Oh, yeah for sure! I'll kick that one up to 5 points to leave breathing room for this and any other things that come up.

PR #119, opened 2023-04-10
clean: remove stackadapt_deprecated

emily-flambe:
Pretty straightforward change here, just suggesting that we nuke all the models for the deprecated stackadapt endpoint from orbit.

@jitoquinto and @kanelouise, curious to hear your thoughts on this - is it better to leave the models in here (not like they're hurting anybody), or remove them so there's no confusion about what the deal is with StackAdapt? I find it hard to imagine anyone will be coming to this repo to look for dbt for a deprecated endpoint 🤔 I am open to opinions!


jitoquinto (approved):
Yeah I think if they're not in use anymore probably better to just yeet them

kanelouise:
++ to the yeeting!

kanelouise (approved):

PR #122, opened 2023-04-11
Add view for pipeline health dashboard

jackwelty:


emily-flambe:
> Per JIRA, it would be slick to have this in a proper dbt project that we orchestrate with Composer, but that can be in a follow up PR! Siiiiiick

Or I mean you can make it be this PR too if you want!

jackwelty:
Going to go ahead and merge this just to keep the record of my original reporting table, but the dbt-ifying will be handled in #123

PR #126, opened 2023-04-24
feat: create dbt models for Empower sync (DE-40)

emily-flambe:
Closes [DE-40](https://techallies.atlassian.net/browse/DE-40)

👏 this 👏  is 👏  working 👏  in 👏  dev 👏  and 👏  prod 👏 

(or at least locally in dev... seems like the DAG is taking its sweet time in dev and twerking just fine in prod, idk)

Roughly a trillion file diffs here so don't expect the closest review but @jitoquinto @kanelouise whichever of you gets to this first, pls to have a look and approve (or don't) 👀 

[DE-40]: https://techallies.atlassian.net/browse/DE-40?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt-cta/empower/models/1_cta_full_refresh/ctaResults_answerIdsByPromptId_base.sql
```diff
@@ -0,0 +1,31 @@
+{% set partitions_to_replace = [
+    'timestamp_trunc(current_timestamp, day)',
+    'timestamp_trunc(timestamp_sub(current_timestamp, interval 1 day), day)'
+] %}
+{{ config(
```
jitoquinto, line 5: All of the `1_cta_full_refresh` models should have `partitions = partitions_to_replace` in the config

kanelouise: This is my only flag! Change this and I believe it shoule be ok.

emily-flambe: ah!! thanks for flagging y'all - I wen ahead and redid these using Jesus' script and I think these are all fixed now!!

dbt-cta/empower/models/1_cta_full_refresh/ctaResults_answerIdsByPromptId_base.sql
```diff
@@ -0,0 +1,31 @@
...
+    cluster_by = "_airbyte_emitted_at",
+    partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
+    tags = [ "nested" ]
+) }}
+-- Final base SQL model
+-- depends_on: {{ ref('ctaresults_answeridsbypromptid_ab3') }}
+select
+    _airbyte_ctaresults_hashid,
+    _13805,
```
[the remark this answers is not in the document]
emily-flambe: yeah I'm a little concerned about these as well... They are from nested fields, so I think if we implement this for a second partner and the fields all have different names, we can A) stop delivering tables from nested fields, and B) go back to the drawing board to figure out the best way to pull out column names for different partners. Not sure the best way to go about that, but I'm willing to kick that can down the road 🙃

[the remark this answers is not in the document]
jitoquinto: sounds good, yeah im kinda leaning towards Option A! a lot of the nested breakouts dont really make sense anyway

jitoquinto (approved):

PR #129, opened 2023-04-25
suggested changes to dbt_format_utils.py

emily-flambe:
@jitoquinto just a couple more suggestions after taking your latest code for a drive, I think once these are changed we should be good to go!!

1) look for base tables ending in SQL - in my case I was hitting errors creating matviews because the script was trying to do it for .DS_Store 🫠
2) I do think we want to keep hashids in the final matviews so that they're present as a default unique row identifier!


jitoquinto:
thanks! :catjam:

PR #131, opened 2023-04-28
feat: WIP script to generate dbt project from a BQ dataset [DE-296]

emily-flambe:
I was feeling spicy so went ahead and got started on some code to generate dbt from raw data living within a BQ dataset. It's pretty inflexible and makes a lot of assumptions, and still a WIP, but definitely does something!

[JIRA](https://techallies.atlassian.net/browse/DE-296)

### DIFF
--- a/utils/generate_dbt/Pipfile
+++ b/utils/generate_dbt/Pipfile
@@ -1,0 +1,14 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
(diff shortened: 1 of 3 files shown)


emily-flambe:
@jitoquinto @kanelouise I'm gonna go ahead and YOLO-merge this since it doesn't impact anything we have going on, just some helper code I wrote for myself that might come in handy down the road. Happy to discuss though if y'all have thoughts!

PR #135, opened 2023-04-30
[DE-294] airbyte_ab_id is not always unique

jackwelty:
This pull request removes the "unique" test from the auto-applied "universal tests" in our schema generation script, since the `_airbyte_ab_id` column is not always unique.

### DIFF
--- a/utils/universal_tests.yml
+++ b/utils/universal_tests.yml
@@ -5,7 +5,6 @@
 columns:
   - name: _airbyte_ab_id
     tests:
-      - unique
       - not_null
   - name: _airbyte_normalized_at
     tests:


jackwelty:
ok @emily-flambe this PR answers the very narrow scope of the ticket. I _think_ I have an idea for how to still allow this test but not for nested fields, but want to run it by you because it's pretty bespoke. 

In the schema gen script, I think we could check to see if a given model has two fields with `hashid` in their name, as this would indicate that we're unpacking a nested field and that `_airbyte_ab_id` should not be expected to be unique any longer. I could hardcode this exception into `generate_schema_yml.py`. I wanted to check before I implemented because that seems pretty niche, and makes this no longer a generic "schema generation" script. Let me know what you think!

emily-flambe (approved):
RIP _airbyte_ab_id uniqueness tests 🥲

I wonder if there's a way to templatize universal_tests.yml so that, eg, we could dynamically apply the uniqueness test except when a field is a nested field, and also add the ability to auto-generate not_null and unique tests for the hashid fields generated in the _ab2 CTEs. idk if that sounds unhinged, don't worry about it right now I'll write a ticket 😛

emily-flambe:
> RIP _airbyte_ab_id uniqueness tests 🥲
> 
> I wonder if there's a way to templatize universal_tests.yml so that, eg, we could dynamically apply the uniqueness test except when a field is a nested field, and also add the ability to auto-generate not_null and unique tests for the hashid fields generated in the _ab2 CTEs. idk if that sounds unhinged, don't worry about it right now I'll write a ticket 😛

Oh I'm being dumb, the change wouldn't be to the universal tests YAML. Don't mind me, this looks good, still waking up over here

PR #136, opened 2023-05-02
[DE-296] POC of pipeline health metrics for non-dbt syncs

jackwelty:
This PR adds a set of dbt models in our `monitoring/` project to parse the output of a GCP Log Sink, configured to output a small subset of logs from Composer. The scope of this ticket just called for a POC implementing our absolute base-level metrics, so the result is an intermediate table (`monitoring.int_parse_scheduler_logs`).

In future tickets, we can combine this table with our existing elementary-based logging and any additional logging we need for other metrics into a set of comprehensive reporting tables.

Deployment steps:
- [x] Create a log sink narrowed to just prod, and point it at an equivalent dataset in our prod project

### DIFF
--- a/dbt-cta/monitoring/models/0_staging/composer_logs/stg_composer_logs__scheduler.sql
+++ b/dbt-cta/monitoring/models/0_staging/composer_logs/stg_composer_logs__scheduler.sql
@@ -1,22 +1,22 @@
-with source as (select * from {{ source("composer_logs", "airflow_scheduler") }})
+with
+    source as (select * from {{ source("composer_logs", "airflow_scheduler") }}),
+    unpack_nested_fields as (
+        select
+            *,
+            resource.type as resource_type,
+            resource.labels.project_id as project_id,
+            resource.labels.environment_name as environment_name
 
-, unpack_nested_fields as (
+        from source
+    ),
+    rename as (
+        select
+            * except (insertid, timestamp, receivetimestamp),
+            insertid as insert_id,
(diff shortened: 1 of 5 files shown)


dbt-cta/monitoring/models/0_staging/composer_logs/stg_composer_logs__scheduler.sql
```diff
@@ -0,0 +1,22 @@
+with
+    source as (select * from {{ source("composer_logs", "airflow_scheduler") }}),
```
jackwelty, line 2: todo: remove `dbt_monitoring` log @jackwelty

dbt-cta/monitoring/models/1_intermediate/composer_logs/int_parse_scheduler_logs.sql
```diff
@@ -79,6 +79,13 @@ with
 
         from unpack_values
     )
+    , filter_unused_dag as (
+        select * from cast_data_types
+        where dag_id not in ('dbt_monitoring')
```
emily-flambe, line 84:
```suggestion
        where dag_id not in ('airflow_monitoring', 'dbt_monitoring') 
```
This is redundant with our log sink exclusion filter, but can we filter out `airflow_monitoring` here as well so it's cute and consistent?

jackwelty: for sure, done!

PR #137, opened 2023-05-03
[DE-290] Show only most recent status per "invocation type" per run

jackwelty:
This PR adds the concept of an "invocation type" to our Pipeline Health Dashboard tables, and calculates the success or failure of a run based only on the most recent execution of a given "invocation type" per run. These "invocation types" are synonymous with Airflow tasks, and this filtering allows us to mark a run as successful if a given Airflow task failed, but succeeded on re-run.

We now surface both of these metrics in the ["All Pipeline Runs" page](https://lookerstudio.google.com/reporting/9f0a1e78-e9ac-4ed5-9788-fe63f40cceff/page/p_r1600o624c) of the dash, so folks can understand when a run may have failed but succeeded on re-run.

Deployment Steps (post-merge):
 - [ ] Run on prod
 - [ ] Update prod data sources
 - [ ] Update/add new column in prod dash
 - [ ] Drop `int_group_by_invocation_status_type` table, no longer used


dbt-cta/monitoring/models/0_staging/elementary/_elementary__models.yml
```diff
@@ -43,4 +43,18 @@ models:
...
         tests:
           - not_null
-          - unique
\ No newline at end of file
+          - unique
+      - name: selected
+        description: >
+          A cleaned up value of the "selected" field, replacing characters with
+          underscrores.
+      - name: invocation_type
```
emily-flambe, line 51: LOVE having this field in here 🙌

dbt-cta/monitoring/models/1_intermediate/pipeline_health/_int_pipeline_health__models.yml
```diff
@@ -17,21 +20,26 @@ models:
...
-              values: ["error","success","fail","warn","pass","skipped"]
-      - name: count
+      - name: test_errors
+      - name: total_tests
+      - name: model_errors
+      - name: total_models
+      - name: most_recent_invocation_type
+        description: >
+          This is a calculated windowed field that tells us which row is the most recent
+          run of an invocation type for a given sync run. This is used to fitler out
```
emily-flambe, line 30:
```suggestion
          run of an invocation type for a given sync run. This is used to filter out
```
lol

dbt-cta/monitoring/models/0_staging/elementary/stg_elementary__dbt_invocations.sql
```diff
@@ -51,6 +52,23 @@ with
             {{ dbt_date.convert_timezone("generated_at", 'America/New_York', 'UTC') }} as generated_at_et
         from extract_vars
     ),
+    add_invocation_type as (
+        select *
+        ,  target_name || '_' || command || '_' || selected as invocation_type
```
emily-flambe, line 57: ohhh this is clever, I like it!!

emily-flambe:
@jackwelty the BQ views in dev aren't reflecting the shiny new things in this PR and the dev PHD is sad (probably as a result), so I added the `dev_deploy` label to see if that updates things in dev 👀

emily-flambe:
@jackwelty this logic all looks sound to me, feel free to merge!!! Two things (non blocking wrt this PR):

1 - The "All Pipeline Runs" page in the PHD is broken tho - ```BigQuery error: Name num_steps_run not found inside t0 at [6:91]``` can you fix that and update here so I can have a look at the goods? 
2 - I just noticed that the `job_*` fields are null - do we know if that's expected or something we should be concerned about? Empirically it hasn't seemed to cramp our style so far, and if it's actually not necessary, let's maybe comment them out of the dbt models and add a comment about what they're supposed to be and why we DGAF?

image

emily-flambe (dismissed):

dbt-cta/monitoring/models/1_intermediate/pipeline_health/_int_pipeline_health__models.yml
jackwelty: spelling is hard!

jackwelty:
> @jackwelty this logic all looks sound to me, feel free to merge!!! Two things (non blocking wrt this PR):
> 
> 1 - The "All Pipeline Runs" page in the PHD is broken tho - `BigQuery error: Name num_steps_run not found inside t0 at [6:91]` can you fix that and update here so I can have a look at the goods? 2 - I just noticed that the `job_*` fields are null - do we know if that's expected or something we should be concerned about? Empirically it hasn't seemed to cramp our style so far, and if it's actually not necessary, let's maybe comment them out of the dbt models and add a comment about what they're supposed to be and why we DGAF?
> 
> image

The job stuff being null is expected! That's from elementary, I think if we were using dbt cloud they would be filled out. I will add a comment.

Re: the PHD, the dev dashboard should be fixed!

PR #138, opened 2023-05-04
feat (ActBlue): implement basic dbt tests 

emily-flambe:
Assigning this to both y'all @jitoquinto @kanelouise, whoever gets around to this first 🙏 

In addition to adding tests, I also added the `_airbyte_*` fields to the materialized views. We are getting requests from partners to include sync metadata, so we should be sure to include those in matviews going forward.

TODO after merging:

- [ ] run dbt locally with `--full_refresh` to rebuild the partner materialized views


dbt-cta/dbt_project.yml
```diff
@@ -41,6 +41,7 @@ models:
         - partner
       +materialized: materialized_view
       +auto_refresh: true
+      +full_refresh: true
```
jitoquinto, line 44: referring [to this other PR](https://github.com/community-tech-alliance/airflow-dags/pull/264) - but i dont think this should be hard coded to true

emily-flambe: agreed! fixed

jitoquinto (approved):

PR #139, opened 2023-05-09
[DE-324] Merge Composer and Elementary Monitoring

jackwelty:
This PR adds a model (`int_merge_logs`) to combine the high-level DAG logging coming from our newly-established log syncs and our existing Elementary logging. 

Putting this PR up for review but flagging something we'll need to address: we don't currently have a way to know which partners are associated with non-Elementary/dbt syncs. All we're extracting is the `dag_id` which doesn't contain this info, so we might need to think about:
1. grabbing more logs
2. checking the information schema? for some of our "rows updated", etc. metrics I think we'll need to log some information schema bits regardless, maybe we could check to see what tables are being updated that match a naming pattern.
3. pulling in some sort of reference/lookup table


emily-flambe:
> we don't currently have a way to know which partners are associated with non-Elementary/dbt syncs

Oh... this is an interesting question @jackwelty, and I'd love to get thoughts from @kanelouise and @jitoquinto about how we can track this. We really need some representation somewhere of vendors, partners, and dag_ids - like, a Google Sheet (which we could read into BQ using a container script)? Another config YAML saved in our airflow repo (version control is nice)? This is actually super important for us to have so that when DAGs fail, we can very quickly communicate about which vendors/users are impacted.

As we think about what the best long-term solution is, @jackwelty what do you think would be the most straightforward POC? And do we need any info besides dag_id, vendor, and partner (maybe also "notes" for any qualitative bits that would be useful to... note)?

jackwelty:
> > we don't currently have a way to know which partners are associated with non-Elementary/dbt syncs
> 
> Oh... this is an interesting question @jackwelty, and I'd love to get thoughts from @kanelouise and @jitoquinto about how we can track this. We really need some representation somewhere of vendors, partners, and dag_ids - like, a Google Sheet (which we could read into BQ using a container script)? Another config YAML saved in our airflow repo (version control is nice)? This is actually super important for us to have so that when DAGs fail, we can very quickly communicate about which vendors/users are impacted.
> 
> As we think about what the best long-term solution is, @jackwelty what do you think would be the most straightforward POC? And do we need any info besides dag_id, vendor, and partner (maybe also "notes" for any qualitative bits that would be useful to... note)?

If we're going for lowest-effort POC, a sheet feels easiest? We can even create it as an [external table](https://github.com/dbt-labs/dbt-external-tables) so no container-scripting required!

emily-flambe:
> > > we don't currently have a way to know which partners are associated with non-Elementary/dbt syncs
> > 
> > 
> > Oh... this is an interesting question @jackwelty, and I'd love to get thoughts from @kanelouise and @jitoquinto about how we can track this. We really need some representation somewhere of vendors, partners, and dag_ids - like, a Google Sheet (which we could read into BQ using a container script)? Another config YAML saved in our airflow repo (version control is nice)? This is actually super important for us to have so that when DAGs fail, we can very quickly communicate about which vendors/users are impacted.
> > As we think about what the best long-term solution is, @jackwelty what do you think would be the most straightforward POC? And do we need any info besides dag_id, vendor, and partner (maybe also "notes" for any qualitative bits that would be useful to... note)?
> 
> If we're going for lowest-effort POC, a sheet feels easiest? We can even create it as an [external table](https://github.com/dbt-labs/dbt-external-tables) so no container-scripting required!

oooh!!! yeah that sounds good! I wrote you a ticket linking to a google sheet to start with. meanwhile this LGTM (approval incoming) 🎉

emily-flambe (dismissed):

PR #140, opened 2023-05-10
Revert "[DE-324] Merge Composer and Elementary Monitoring"

emily-flambe:
Reverts community-tech-alliance/dbt-cta#139

@jackwelty we are getting errors in prod so I am rolling this back for now!


jackwelty:
I can pick this up next week but I figured out the issue @emily-flambe !

Basically if we re-run tasks in a DAG, then we trigger a new `DagRun Finished` log, with new run start / ends, but with the same `dag_id` and `run_id`. We'll just have to add another intermediate step that groups by `run_id`, and probably takes a min/max start/end time.

Couple examples here (in prod):
```sql
select * from `monitoring.stg_composer_logs__scheduler`
where insert_id in ('kjadf7ff7tjyo','1de34ugf8a4m8w');
```

PR #143, opened 2023-05-11
feat: set auto_refresh to `false` for partner matviews

emily-flambe:


dbt-cta/dbt_project.yml
```diff
@@ -40,7 +40,8 @@ models:
       +tags: 
         - partner
       +materialized: materialized_view
-      +auto_refresh: true
+      +auto_refresh: false
+      +full_refresh: true
```
jitoquinto, line 44: This will drop and rebuild matviews right? is that what we want to be doing?

emily-flambe: yep - we want the refresh to happen as part of the dag run so that the "last modified" metadata in BQ reflects the timing of the sync. And with auto-refresh enabled, refreshing at sync-time means MVs are getting refreshed twice, which is 2x the compute cost we need. It seems like dropping-and-rebuilding matviews doesn't break any views that are built on top of them, so it shouldn't have any ill effects

jitoquinto (dismissed):

dbt-cta/dbt_project.yml
emily-flambe: per Slack, just updated full_refresh to be explicitly false!

michaelefisher (approved):

PR #148, opened 2023-05-14
[DE-250] deliver matviews (not tables) for Stripe data

emily-flambe:
This PR updates the models in the `stripe` sync to deliver materialized views instead of tables.

## materialized views
These are all defined using SELECT * statements, so I just renamed the folder from `2_partner_tables` to `2_partner_matviews`. To actually convert them, I ran dbt locally with the `--full-refresh` option (see below).

## base models
Stripe has WAY too many models for me to clicky-copy-pasty all of this manually, so I created a script ([utils/convert_to_full_refresh.py](https://github.com/community-tech-alliance/dbt-cta/blob/df51194d13b5326511376422f84062b6b38d1d0c/utils/convert_to_full_refresh.py)) to add the partition replacement logic. The base models are definitely worth a closer look to make sure I didn't mess this up. The script might come in handy when doing this for other syncs.

## Testing

dbt commands all succeeded when testing locally. Dev Composer seems to give up when running the CTA models (this was true before deploying the changes in this PR) - possibly need to bump the CPU or worker limits or something 🤔 

Note that to convert the tables to materialized views, we need to run the dbt locally with the `--full-refresh` option added:

```
pipenv run dbt run --target partner --select tag:partner --full-refresh
```

(I updated our [Cacher runbook](https://snippets.cacher.io/snippet/1db187592eb6ab37ab0b) to clarify this step)

I've already run that command in dev and will need to do the same in prod once this is merged.

## TODOs (post merge):

- [ ] run prod partner models locally with `--full-refresh` to convert tables to materialized views


emily-flambe:
@kanelouise @jitoquinto bumping this too since there are a lot of PRs flying around 😵‍💫

dbt-cta/stripe/models/1_cta_full_refresh/balance_transactions_fee_details_base.sql
```diff
@@ -1,3 +1,13 @@
+{% set partitions_to_replace = [
```
jitoquinto, line 1: looks like this `set partitions_to_replace` block of code is in here twice for all the base models

dbt-cta/stripe/models/1_cta_full_refresh/subscription_items_base.sql
jitoquinto, line 1: This model is missing the `is_incremental` block

jitoquinto (changes requested):
Left some comments but it looks like all the base models have an extra `set partitions_to_replace` block. I also found one model with a missing `in_incremental` block at the end. I just kinda did a quick scroll through tho so there might be more

dbt-cta/stripe/models/1_cta_full_refresh/balance_transactions_fee_details_base.sql
emily-flambe: Oh!! yep I know exactly why that happened... fixing now

kanelouise: Seems like this is fixed now. If it passes dbt tests locally, this has my stamp of approval.

emily-flambe:
ok @jitoquinto - sorry about the careless mistakes. The `set_partitions_to_replace` blocks appeared twice because I had run the script twice 🤦‍♀️ and the logic I had in the script was only updating the filter for:

```
where 1 = 1
{{ incremental_clause('_airbyte_emitted_at') }}
```

which meant that the 10ish base models that didn't have the incremental_clause in there weren't having that filter updated. 

I THINK the changes I just pushed will do the trick. I'll have a scroll through the diff now to see if anything is obviously amiss...

kanelouise (approved):
With your most recent changes, @emily-flambe, I don't see any issues.

emily-flambe:
> With your most recent changes, @emily-flambe, I don't see any issues.

Ty!! I'm still gonna hold off until after tomorrow's sync to merge this in, just in case 😶‍🌫️

PR #149, opened 2023-05-15
un-reverting "feat: set auto_refresh to `false` for partner matviews"

emily-flambe:
now that it's Monday let's give this a go once the syncs are all done (one remaining)

### DIFF
--- a/dbt-cta/dbt_project.yml
+++ b/dbt-cta/dbt_project.yml
@@ -40,7 +40,7 @@ models:
       +tags: 
         - partner
       +materialized: materialized_view
-      +auto_refresh: true
+      +auto_refresh: false
       +refresh_interval_minutes: 1440
       +grant_access_to:
         - project: "{{ env_var('CTA_PROJECT_ID') }}"


emily-flambe:
hmmm @michaelefisher so this didn't actually have the intended effect of disabling auto-refresh on partner materialized views. I'm going to do a little science with the mailchimp sync to locally runn the dbt with --full-refresh (to disable auto-refresh) and then run the dbt in Composer to see if auto-refresh remains `false` thereafter 🧐

michaelefisher:
@emily-flambe I wonder if dbt isn't issuing the correct API request, which is to say maybe it isn't updating the view because it already exists? We could probably check logs in Cloud Logging to see what dbt _tried_ to do to the view (if anything).

emily-flambe:
ok even with the updated dbt_project.yml file, subsequent runs of partner dbt in Composer aren't disabling auto-refresh on materialized views.

@michaelefisher I'm inclined to just rerun all the partner matviews locally with `--full-refresh` rather than poke around in the logs to understand exactly why this isn't working... Once we disable auto-refresh then the dbt in Composer shouldn't cause that status to change. (I'll do some science first...)

emily-flambe:
Wait, now I'm properly confused - rerunning the dbt in Composer did not disable auto-refresh for mailchimp, but it seems like it did for empower? please hold, science is ongoing (maybe Composer was using a cached version of dbt_project.yml... it happens)

PR #154, opened 2023-05-22
[DE-338] Union Composer and Elementary logs into final table for PHD

jackwelty:
This PR makes edits to our final reporting table for the PHD to incorporate our changes merging Composer and Elementary logs.

The `rpt_pipeline_health` table now:
* Uses `int_merge_logs` as a base (our merged Composer + Elementary logs)
* Left joins in Elementary/dbt-specific metrics (# of models run, # of tests, etc.)
  * Our non-DMDF syncs will not have these metrics but that's ok, those columns/charts will just be null in the PHD.
* Computes status across those two sources, flagging a failure if it appears in either dataset. We use the `status` field so that every DAG run has a status, regardless of which logging source we're using

This structure means we can upload the general structure of the PHD, with some minor changes to use our new `status` field, surface `dag_id`, etc. 

TODO (post-merge):
 - [ ] Turn dbt-monitoring DAG back on in dev
 - [ ] Update prod dashboard


jitoquinto (dismissed):
:catjam:

emily-flambe (approved):
I am loving all the TESTS 🎉  let's gooo

PR #155, opened 2023-05-22
feat (google_directory): convert 1_cta_tables to 1_cta_full_refresh [DE-245]

emily-flambe:
This PR modifies the google_directory models so that they use 1_cta_full_refresh instead of 1_cta_tables. This sync is currently for internal use only, so there are no partner models to convert to materialized views 🙌 

~Because this change adds partitions to tables that were not previously partitioned (and I've already run this dbt in prod), we will need to merge this change before tomorrow's sync or else it will fail - @kanelouise @jitoquinto could I get some eyes on this today? 🙏🙏🙏 should be fairly straightforward, not that many tables here!~ actually the sync succeeded just fine, so my spook was unfounded - still would love eyes on this to close it out @jitoquinto @kanelouise 

TODO (post-merge):

- [ ] drop and rebuild tables so that they have partitions
- [ ] confirm dbt with partition replacement logic runs successfully


dbt-cta/google_directory/models/1_cta_full_refresh/group_members.sql
```diff
@@ -0,0 +1,26 @@
+{% set partitions_to_replace = [
+    'timestamp_trunc(current_timestamp, day)',
+    'timestamp_trunc(timestamp_sub(current_timestamp, interval 1 day), day)'
+] %}
+
+{{ config(
+    cluster_by = "_airbyte_emitted_at",
```
jitoquinto, line 7: These partition replacement tables need the following config added too!

```
partitions = partitions_to_replace,
```

emily-flambe: oh dang, those are missing from the stripe models too! I'll make a separate PR to add those in...

emily-flambe: OK, just added `partitions=partitions_to_replace` in all the google_directory CTA base model configs!

jitoquinto: sweet thanks!

jitoquinto (approved):

PR #158, opened 2023-05-23
Fix: relax nullness threshold in monitoring pipeline

jackwelty:
We expect `partner_name` and `sync_name` to not be null, because for our Composer-logged-only syncs (i.e. not dbt/DMDF), we join in a metadata sheet. We have a dag ID (`targetsmart_xrefs_sync`) that is not in the sheet, and our nullness rate is hovering right around 5% and occasionally failing the DAG. We should document this sync in our metadata sheet but also bump this threshold.


emily-flambe (dismissed):
I don't wanna block merging this, would be great if we could add commentary around this test (just the same explanation you gave in slack more or less)! Merge whenever tho!

PR #160, opened 2023-05-23
feat (stripe): add `partitions=partitions_to_replace` to 1_cta_full_refresh configs

emily-flambe:
these were missing the first time around... oops

@jitoquinto


jitoquinto (approved):
oof good catch!

PR #164, opened 2023-05-26
hotfix: use `source` instead of `ref` for partner matviews

emily-flambe:


emily-flambe:
@kanelouise (cc @jitoquinto) cool fact, I did a Big Goof when recommending we use `ref` for table references in partner matview definitions! I had forgotten that we need to use `source` because the partner matviews are referencing tables that live in a separate project. `ref` works only when the tables are all in the same project and dataset, which I'm guessing was the case when you ran locally.

This PR rights the wrongs of my silly recommendation 🥲 

I guess there is a lesson here for us all 🌈

jackwelty:
> @kanelouise (cc @jitoquinto) cool fact, I did a Big Goof when recommending we use `ref` for table references in partner matview definitions! I had forgotten that we need to use `source` because the partner matviews are referencing tables that live in a separate project. `ref` works only when the tables are all in the same project and dataset, which I'm guessing was the case when you ran locally.
> 
> This PR rights the wrongs of my silly recommendation 🥲
> 
> I guess there is a lesson here for us all 🌈

I'm very late here but if in the future we want to try this again, we could specify a [custom schema](https://docs.getdbt.com/docs/build/custom-schemas) for just the matviews. It would likely involve some changes to the `generate_schema_name()` macro, but the refactor might be worth it to simplify our source/target structure.

PR #165, opened 2023-05-27
clean: remove models from dbt-project.yml (no longer needed)

emily-flambe:
@jitoquinto @kanelouise Now that we have FINALLY completed our quest of only delivering materialized views to partners (not tables), the time has come to remove these options from dbt-project.yml, so that we may never again stray down that dark path.

Also clarified the commentary surrounding the special model types for `google_ads` and removed one that we don't actually need. 🧹 ✨ 🧹 ✨ 🧹 ✨

### DIFF
--- a/dbt-cta/dbt_project.yml
+++ b/dbt-cta/dbt_project.yml
@@ -27,10 +27,6 @@ models:
       +incremental_strategy: insert_overwrite
       +on_schema_change: sync_all_columns
       +full_refresh: false
-    1_cta_tables:
-      +tags:
-        - cta
-      +materialized: table
     1_cta_incremental:
       +tags:
         - cta
@@ -45,10 +41,6 @@ models:
       +grant_access_to:
         - project: "{{ env_var('CTA_PROJECT_ID') }}"
           dataset: "{{ env_var('CTA_DATASET_ID') }}"
-    2_partner_tables:
-      +tags: 
-        - partner
-      +materialized: table
     3_partner_views:
       +tags: 
         - partner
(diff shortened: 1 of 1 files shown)


michaelefisher:
Before we do this, can we confirm that we definitely don't want tables anymore (see bug here: https://techallies.atlassian.net/browse/DE-362?atlOrigin=eyJpIjoiY2I2YjY2NGQyZTA0NGEyZWJkMmNmMmFhNzI3ZmQxNzciLCJwIjoiaiJ9)

emily-flambe:
> Before we do this, can we confirm that we definitely don't want tables anymore (see bug here: https://techallies.atlassian.net/browse/DE-362?atlOrigin=eyJpIjoiY2I2YjY2NGQyZTA0NGEyZWJkMmNmMmFhNzI3ZmQxNzciLCJwIjoiaiJ9)

We just spent several months working through tech debt tickets to deliver matviews instead of tables. The goal of our entire dbt setup is to perform any deduping (etc) logic so that the content of the `_base` tables exactly match what we deliver to partners in materialized views, which should just be SELECT * statements (unless we care about omitting `_airbyte_*` fields). I feel pretty strongly that we should keep this methodology consistent across all of our partner syncs, even if it requires additional transformation steps to deliver the `_base` tables the way we need them.

emily-flambe:
bumping this for 👀 @kanelouise @jitoquinto

jitoquinto (approved):
Im good with this as long as we dont have any models still using these!

PR #167, opened 2023-06-01
feat: make facebook_marketing full_refresh

kanelouise:
Per [this ticket](https://techallies.atlassian.net/browse/DE-362)

This PR simply updates `1_cta_full_refresh` to match our standard full refresh standard. We are switching to a full refresh strat because the incremental strat failed a few times last week when there were duplicate records in the delivered tables. 

The dbt ran locally just fine, as did the dbt tests.


emily-flambe (approved):
This seems fine to me! Curious @jackwelty if you see any issues with switching to this approach?

emily-flambe (approved):

PR #172, opened 2023-06-05
[DE-316] Add POC volume anomaly test

jackwelty:
This is a POC implementation of a volume anomaly test, to see if the generated metadata would help fulfill the ACs for DE-316. It is not meant to be merged.

Note: there is a bug somewhere between `elementary` and `dbt_utils` that is causing elementary's `current_timestamp()` macros to return `current_timestamp::timestamp`, which contained a cast expression (`::`) that is not valid in BigQuery. I've asked a question about this in elementary Slack, but in the meantime I was able to resolve this by overwriting the `edr_current_timestamp()` and `edr_current_timestamp_in_utc()` macros in `dbt-cta/dbt-packages/elementary/macros/utils/cross_db_utils/current_timestamp.sql` as follows:

```sql

{% macro edr_current_timestamp() -%}
    current_timestamp
{%- endmacro %}

{% macro edr_current_timestamp_in_utc() -%}
    current_timestamp
{%- endmacro %}
```
Because this is an edit to macros contained in a dbt package, I can't check this into git. It might be worth figuring out how we override them somehow.

This _might_ have something to do with the fact that we have our own `macros/cross_db_utils/`, which appears to be a subset (?) of what the `dbt_utils` package.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 requests = "*"
 dbt-core = "*"
-dbt-bigquery = "==1.3.0b2"
+dbt-bigquery = "==1.3.0"
 gcloud = "*"
 pyyaml = "*"
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "cf536953cf81ecf16b379cee8fd34954164577c775f1360a9029363883268287"
+            "sha256": "010cc28d6ee7ed42914fa0547aa2f771230abde6b911df3add4cd417696de7b0"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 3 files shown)


emily-flambe:
@jackwelty wow this is some heroically hacky antics! Can you say more about why this can't be checked into git? Like, does your hackystrat work when running locally, but it wouldn't actually work in our composer-run dbt?

jackwelty:
> @jackwelty wow this is some heroically hacky antics! Can you say more about why this can't be checked into git? Like, does your hackystrat work when running locally, but it wouldn't actually work in our composer-run dbt?

These changes are actually made inside a macros file that is part of the `elementary` dbt package. We `dbt-cta/dbt_packages/` is (correctly) part of our `.gitignore` because that's all 3rd party code that we install via `dbt deps` (and then copy to Composer, but that's mostly just an artifact of our current deployment workflow). I think finding out a way to override these is likely the easiest path forward.

jackwelty:
I figured it out!! We were running a _slightly_ outdated version of `dbt-bigquery` than we should have been `1.3.0b2` (some sort of beta release?) instead of `1.3.0` (the normie, prod 1.3 release), which did not included the `bigquery__current_timestamp_backcompat()` macro. 

TL;DR if we want to use these tests (and I think we do) then we actually do need to merge this PR and bump our dbt-bq version ever so slightly. In this moment I don't exactly remember what the process of updating Airflow PyPI dependencies is but I remember it's a little delicate.

PR #179, opened 2023-06-13
feat (NEW SYNC): add dbt models and tests for Hustle [DE-61]

emily-flambe:
# Summary

This dbt uses raw data delivered by our (shiny new) Hustle S3-to-BQ sync and delivers the data to partners via materialized views with rows deduped by a primary key (or hashid where no PK exists).

This PR is meant to be reviewed and deployed in tandem with [this PR in the airflow-dags repo](https://github.com/community-tech-alliance/airflow-dags/pull/309), which adds the dbt tasks to the DAG.

## row deduplication

Deduping happens in `*_cte2` for each table. This ensures that only one row exists for each primary key value based on whichever one is most recent (based on `updated_at` where possible, or `_cta_sync_datetime_utc` otherwise).

## incremental strategy

The source data update incrementally, so all base table models use `1_cta_incremental`, merging in new rows based on the unique key (usually `id`, sometimes `_cta_hashid`, see below).

## format

All models for the Hustle sync use a consistent format, with two possible variations:

- tables with no `id` dedupe using `_cta_hashid` as the primary key, which is generated as a hash of all columns in the table (except for `_cta_sync_datetime_utc`, which is added in `*_cte1`)
- tables with no `updated_at` dedupe using `_cta_sync_datetime_utc` as the field for identifying the most recent row of data for each primary key.
- (both of the above might be true for a table, which would then dedupe using the most recent `_cta_sync_datetime_utc` for each `_cta_hashid`)

## TODOs

Before marking this ready for review:

[x] - add some row uniqueness tests. I think this should be fast?! ✅ 🙌 
[x] - modify the DAG to add dbt tasks
[x] - confirm dbt works when running in Composer
[x] - I'm gonna make a LOOM for this PR! Oh boy! [DONE](https://www.loom.com/share/bd3baa58a190443ba7475097cd4a341b?sid=1cf4d756-eaf2-4f17-bd37-9bdef4754159)


emily-flambe:
@jitoquinto @kanelouise @royconst this PR is ready for review! and it has a [LOOM](https://www.loom.com/share/bd3baa58a190443ba7475097cd4a341b?sid=1cf4d756-eaf2-4f17-bd37-9bdef4754159)! lmk if that ends up being useful 🧐 

also gonna mark its sibling-PR in airflow-dags ready for review as well!

PR #185, opened 2023-06-20
feat (freshdesk): create basic dbt tests [DE-210]

emily-flambe:
As recorded by this three-part Loom video series (for CTA eyes only, sorry looky-loos!)

[Part 1](https://www.loom.com/share/02ced0dc409048b49643fa1b640fdbfb?sid=f43e3d47-6f99-4883-a13e-a7ebc59cd542)
[Part 2](https://www.loom.com/share/c3c8cf6744864ce28d187555bc9b0ee5?sid=9b2f706c-303f-49b8-99f1-394b22d3b802)
[Part 3](https://www.loom.com/share/e2284d333f0644c391fefee11980ddaf?sid=c3d75a76-b7a8-42d6-a40b-8524c36279e7)

This PR implements the following dbt tests on the Freshdesk base tables:

- `_airbyte_emitted_at` should be a timestamp
- `_airbyte_ab_id` should be unique and not null (this should be true if 1_cta_full_refresh is replacing date partitions as expected, meaning that `min(_airbyte_emitted_at)` should never be smaller than yesterday's date)
- `id` should be unique and not null - I added this using my _intuition_ and it seems to hold up for everything except the table `survey_questions_base`, for which `id` is a string value like "default question," and clearly not a proper UPK.

These tests are passing in both dev and prod (when running locally) after having deleted straggler rows that had `_airbyte_emitted_at` values back in April - I think this happens when the sync gets temporarily messed up, because the models in 1_cta_full_refresh only rewrite today's and yesterday's date partitions, so if there's a longer interval between successive runs, an old partition might remain in the base data. These tests will fail (as they should) if that happens again in the future.

### DIFF
--- a/dbt-cta/freshdesk/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/freshdesk/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -7,6 +7,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: agent_id
         description: ''
       - name: ticket_id
@@ -45,6 +48,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: name
         description: ''
       - name: description
@@ -71,6 +77,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: name
         description: ''
       - name: description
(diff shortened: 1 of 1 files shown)


jitoquinto (approved):
🔥 thanks for the looms!

jitoquinto:
oh one question, the universal tests has tests for the `airbyte_normalized_at` column. Should these have that test too?

emily-flambe:
> oh one question, the universal tests has tests for the `airbyte_normalized_at` column. Should these have that test too?

Good catch! We probably should actually drop that from universal_tests.yml to avoid confusion since we don't actually use `_airbyte_normalized_at` - for partners that care about metadata surrounding when data were synced, we will just be pointing them to `_airbyte_emitted_at`.

PR #186, opened 2023-06-22
fix: dbt tests generator add universal tests to all models

jitoquinto:
https://techallies.atlassian.net/browse/DE-410

Looks like the issue was just  `model_config in model_configs[0]:` It should've been accessing `model_config[0]`. 
I also did a bit of renaming and breaking out since some of the code was hard for me to understand haha


emily-flambe (approved):
Awww yeah! Way to leave the campsite better than you found it 🎉 I tested this with usvote_foundation and it seems to be doing the thing, so you're good to merge this and close out the ticket!!

PR #188, opened 2023-06-28
hotfix (freshdesk): dedupe on `id` for tables `conversation` and `ticket`

emily-flambe:
dbt tests for `id` uniqueness have started failing for `conversation` and `ticket`. These appear to be unwanted dupes originating in the source data, so this PR introduces CTEs that dedupe using only the most recently emitted row for each `id`.


kanelouise (approved):
LGTM!

PR #189, opened 2023-07-03
feat (blocks): add new models and tests [DE-285]

emily-flambe:
While implementing the Blocks sync for a new partner, I found a number of tables that were not included in the original dbt because they were absent from the original data source we were using to develop the sync.

This means that not all partners will use all possible models included in this sync, which in turn means that a partner whose block data is not using some of the models will fail dbt runs (due to missing `_airbyte_raw_*` source tables) until those tables are initialized (with zero rows of course).

Possible TODO(): add that as a step to the DAG, or at least write some kind of script to initialize missing source tables? seems like premature optimization for now, but might be necessary later on in order to preserve our sanity.

(see also the PR that modifies the DMDF config to start delivering this data to the TOT project: https://github.com/community-tech-alliance/airflow-dags/pull/327)


royconst (approved):
Looked at this a couple times. It's a lot of files but don't see anything that stands out as wonky. Don't really have an idea on how to handle the tables but the scenario is interesting. I'll keep it in mind when plugging in dbt in the future.

emily-flambe:
> Looked at this a couple times. It's a lot of files but don't see anything that stands out as wonky. Don't really have an idea on how to handle the tables but the scenario is interesting. I'll keep it in mind when plugging in dbt in the future.

Yeah, this is the first time we've encountered this scenario (same sync with different subsets of tables actually being synced for different partners) - gonna be a cute little technical challenge for us to worry about once we start syncing this data for State Voices in addition to TOT, but we don't _need_ to worry about it until then 😎

jitoquinto:
Just leaving an idea here before it leaves my brain haha. We could try something with the `--exclude`[ flag when running dbt](https://docs.getdbt.com/reference/node-selection/exclude). And it seems like [these dbt selectors](https://docs.getdbt.com/reference/node-selection/yaml-selectors#exclude) could be used to create a group of models to exclude. It seems like we can make some Selectors that exclude or include specific models based on vendor offerings (so depending on a partners deal with a vendor, we can include/exclude a set list of models)

emily-flambe:
> Just leaving an idea here before it leaves my brain haha. We could try something with the `--exclude`[ flag when running dbt](https://docs.getdbt.com/reference/node-selection/exclude). And it seems like [these dbt selectors](https://docs.getdbt.com/reference/node-selection/yaml-selectors#exclude) could be used to create a group of models to exclude. It seems like we can make some Selectors that exclude or include specific models based on vendor offerings (so depending on a partners deal with a vendor, we can include/exclude a set list of models)

I'm just working through messages from my week of PTO lol but this sounds VERY promising!! I'll write myself a reminder to slap it into a ticket... at some point.

PR #191, opened 2023-07-05
feat (google ads): add dbt tests [DE-211]

emily-flambe:
Pretty straightforward since there are only 2 tables we actually deliver in this sync, and the only obvious test to add is to make sure `hashid` is unique. Also added tests for `unique_combination_of_columns` based on the fields that are used to generate the hashid. That _should_ be redundant with the hashid row uniqueness test but I was feeling spicy.

@kanelouise @royconst


royconst (approved):
Looks fine :thumbsup:

PR #192, opened 2023-07-11
fix for an issue with turf tables

royconst:
Moved turf_levels_base.sql to full_refresh. 
Ran the dbt locally and everything was successful. 
I cleared the failed upstream tests on the dag and marked the run successful. All the tests then passed on the dag.
Incidentally it also fixed the permissions errors that we had a prog lead ticket open for.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 requests = "*"
+chardet= "*"
 dbt-core = "*"
 dbt-bigquery = "==1.3.0b2"
 gcloud = "*"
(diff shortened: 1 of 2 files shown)


jitoquinto (approved):
lgtm!

Pipfile
```diff
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 requests = "*"
+chardet= "*"
```
emily-flambe, line 8: This PR was already merged a few days ago and I'm sure this isn't important, but @royconst curious if you know how this made it into the pipfile in dbt-cta 🤔 did this slip in by accident or is it needed for something?

PR #196, opened 2023-07-20
feat: Init Zoho Creator dbt for Metadata portion of sync

jitoquinto:
https://techallies.atlassian.net/browse/DE-429
dbt for the new Zoho Creator connector! Note this is only for the Meta API Endpoints and not the Bulk endpoints

PR for full DAG which is also syncing Bulk Reports - https://github.com/community-tech-alliance/airflow-dags/pull/349


emily-flambe (approved):
As gnarly as the whole Zoho Creator sync is, this at least seems straightforward, and I see no flags :shipit:

kanelouise:
dbt looks good to me, too.

PR #197, opened 2023-07-21
feat: add dbt models for ADP Worker Management [DE-304]

emily-flambe:
Hey, we've got some models here! I even created a real-ass README using most of the PR description, see below.

Sample tables yielded from this dbt can be viewed in dev BQ in dataset `partner_a_adp_worker_management`, and materialized views are in the dev Partner A project in dataset `adp_worker_management`.

---

A note about how this sync works: Airbyte delivers all of the data in a single table, `_airbyte_raw_workers`, in whose field `_airbyte_data` are contained all of the fields in the sync. Most of these are captured in the table `workers`, and some, such as contact info fields (personal_emails, work_emails) are array fields that are unnested in separate models. This is necessary because a single worker (associateOID in `workers`) might have more than one row for those streams.

In addition, the array field `workerAssignments` has two array fields that are further unnested: `assignedOrganizationalUnits` and `homeOrganizationalUnits`.

Beyond the mild unnestery going on, the dbt for this sync is fairly straightforward.

## Tests

This PR includes source tests, which look a lot like the tests we encode in schema yml files (to come in a separate PR) but are actually just included in `sources.yml`. I don't know why I chose to add them here, since in theory they are duplicative of the tests yet to come. I must have been in some type of mood. But if a test is good enough to run once, it's good enough to run twice, I guess?

All tables are updated using the Full Refresh - Overwrite sync mode in Airbyte, and all CTA base models use the full refresh strategy (`1_cta_full_refresh`). As such, the unique primary key for each table is the `_airbyte_*_hashid` generated in the `*_ab3` step in 0_ctes, and the tests are all configured to check for non-null-ness and uniqueness for these `hashid` fields.


jitoquinto (approved):
wowow this breakout is so much cleaner than whatever Airbyte is doing when it breaks out nested fields!

PR #199, opened 2023-07-24
feat (ADP worker mgmt): add schema yaml with dbt tests [DE-441]

emily-flambe:
This PR compares a new branch with dbt tests for the new ADP Worker Management sync to the existing branch (also in review) that adds all the other files to initialize the dbt for that sync.

@kanelouise @jitoquinto @royconst, probably easiest to review the other PR first (for Phase 2, [here](https://github.com/community-tech-alliance/dbt-cta/pull/197)) and then take a glance at this one.

dbt models and tests are all running successfully locally - about to try a dev deploy to airflow 👀


jitoquinto (dismissed):

kanelouise:
The only flag I have on this phase is that the tests are only on AB ID's and not primary keys from ADP itself. Maybe something you'd like to add in the future @emily-flambe?

emily-flambe:
> The only flag I have on this phase is that the tests are only on AB ID's and not primary keys from ADP itself. Maybe something you'd like to add in the future @emily-flambe?

Ahhh thank you for being on the lookout for that @kanelouise !! In this case actually all of the tables except for `workers` come from nested fields, so there's no PK in the data itself - instead we only have the `*_hashid` fields to rely on for avoiding duplicate rows. But I will YOLO-merge in a test for `AssociateOID` uniqueness to `workers_base` because that is the PK for that table!

PR #200, opened 2023-07-24
feat: add mobilize dbt tests

jitoquinto:
@emily-flambe what do we do if we run tests and dev and find dupes 🙃 ?? Should i run in prod and see if its a problem there? 

Anyway, here are the dbt tests for mobilize! this is my first time doing this so let me know if I'm missing anything or if i did something wrong!

### DIFF
--- a/dbt-cta/mobilize/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/mobilize/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -7,6 +7,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: source
         description: ''
       - name: user_id
@@ -71,6 +74,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: status
         description: ''
       - name: user_id
@@ -177,6 +183,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: name
         description: ''
       - name: slug
@@ -209,6 +218,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: end_date
         description: ''
       - name: event_id
@@ -249,6 +261,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: tag_id
         description: ''
       - name: event_id
@@ -287,6 +302,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: van_id
         description: ''
       - name: user_id
@@ -323,6 +341,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: van_id
         description: ''
       - name: event_id
@@ -363,6 +384,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: title
         description: ''
       - name: summary
@@ -547,6 +571,9 @@ models:
     columns:
       - name: id
         description: ''
+        tests:
+          - not_null
+          - unique
       - name: van_id
         description: ''
       - name: end_date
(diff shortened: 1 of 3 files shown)


dbt-cta/mobilize/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,804 @@
+version: 2
+
+models:
+
+  - name: affiliations_base
+    description: ''
+    columns:
+      - name: id
```
kanelouise, line 8: I think it could be useful to add uniqueness tests on the primary keys of these non-ab base tables 🤷

kanelouise (approved):
I found nothing to flag. Looks good, my one suggestion would be to add primary key tests to the base tables that are not an AB ID. Redundant? Maybe. But it could also reveal a dupe where the AB ID is different but the table primary key is duplicated.

dbt-cta/mobilize/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
emily-flambe: agreed! logically the `*_hashid` fields _should_ also function as primary keys, but it seems like a good idea to also test for uniqueness when there's another field that seems like it ought to be the obvious PK for the table (I think generally a safe assumption when it's a field called `id`). Even if those tests should be redundant, it can still be useful to have them both - and if that reveals that an "obvious" UPK is not actually unique, that might be something we document somewhere (maybe the README for the dbt project).

Bottom line, I agree with @kanelouise, let's add non-null and unique tests to all the `id` fields 😎

utils/cta_dbt_helper.sh
```diff
@@ -131,7 +131,7 @@ generate_dbt_tests() {
         INPUT_DIR_NAME=$(gum input --prompt "Enter the name of the vendor to generate tests for: " --placeholder "(ex. actblue)")
         OPTION_UNIVERSAL_TESTS=$(gum input --prompt "Would you like to initialize these files using the default tests (see universal_tests.yml)? " --placeholder " (Y is recommended! Or leave blank to skip)")
         if [ "$OPTION_UNIVERSAL_TESTS" == "Y" ]; then
-            CLI_OPTIONS="--universal-tests-path ../utils/universal_tests.yml"
```
emily-flambe, line 134: Ooh thank you for making this change, definitely more legit!

emily-flambe (dismissed):
I agree with kane's suggestion of adding uniqueness and not-null tests to the `id` columns for each table, though that's not technically a blocker! Congrats on your first dbt test experience, bonus points awarded for including source tests as well (I sometimes forget about those 🫠

jitoquinto:
> I found nothing to flag. Looks good, my one suggestion would be to add primary key tests to the base tables that are not an AB ID. Redundant? Maybe. But it could also reveal a dupe where the AB ID is different but the table primary key is duplicated.

Thats a good point! I agree that if we can find a primary key that isnt AB generated, we should! def have ran into the case where AB ID was unique but the row itself wasnt 🙃. I'll look through and add PK tests where they seem appropriate!

utils/cta_dbt_helper.sh
jitoquinto: yeah im hoping to make this script not care what directory you run it from!

dbt-cta/mobilize/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
jitoquinto: makes sense! ill give it a go and try to find PKs to add tests to :catsalute:

jitoquinto:
@emily-flambe @kanelouise added those uniqueness and not null tests to `id` fields! I ran the tests for a few of the sv_mobilize_org datasets in prod and they all passed 🎉  so those id's are indeed PKs

emily-flambe:
> @emily-flambe @kanelouise added those uniqueness and not null tests to `id` fields! I ran the tests for a few of the sv_mobilize_org datasets in prod and they all passed 🎉 so those id's are indeed PKs

hell yea 🙌 ship it!

jitoquinto:
> > @emily-flambe @kanelouise added those uniqueness and not null tests to `id` fields! I ran the tests for a few of the sv_mobilize_org datasets in prod and they all passed 🎉 so those id's are indeed PKs
> 
> hell yea 🙌 ship it!

:catjam: @emily-flambe could you re-review?

emily-flambe (approved):
ope!

PR #204, opened 2023-07-28
dbt models for Workflows executions -> PHD [DE-381]

emily-flambe:
I made a log sink

~I'll figure the rest out later lol~

I figured out the rest! Check out the new goods in the dev PHD:

- [overview](https://lookerstudio.google.com/reporting/9f0a1e78-e9ac-4ed5-9788-fe63f40cceff/page/p_afmcsb424c), now with a column for "sync type" (Composer or Workflows)
- and a cute new page with [details about workflow executions](https://lookerstudio.google.com/reporting/9f0a1e78-e9ac-4ed5-9788-fe63f40cceff/page/p_85o1depj8c)!
 
What do we think team? @royconst @kanelouise @jitoquinto 

Note also the [accompanying PR in airflow-dags](https://github.com/community-tech-alliance/airflow-dags/pull/366) that adds the Google Worksheet ID for storing metadata about workflows.


jitoquinto (dismissed):
wowow love all the new PHD additions! So the `Overview` tab will only show workflows in the sheet but the `Workflow Executions` tab shoes _all_ the executions?? also, could you go over how this is working at an eng weekly at some point?? Im having a hard time following whats going on by just reading dbt 😵 and that would be really helpful!

dbt-cta/monitoring/models/0_staging/workflows_execution_logs/stg_workflows_execution_logs.sql
```diff
@@ -0,0 +1,30 @@
+with
+    source as (select * from {{ source("workflows_logs", "executions") }}),
+    unpack_nested_fields as (
+        select
+            * except (logName, resource, textPayload, jsonpayload_type_executionssystemlog),
+            JSON_EXTRACT_SCALAR(jsonpayload_type_executionssystemlog.start.argument,'$.partner_name') as partner_name,
```
jitoquinto, line 6: so cool! gotta go through and add this arg into all our currently running workflows!

emily-flambe: > So the Overview tab will only show workflows in the sheet but the Workflow Executions tab shows all the executions...?

Currently the Overview shows the status of the most recent run per workflow_id * partner per day - it's not clear that's the best representation of "Workflow Sync Status", given that workflows might run at arbitrary times (unlike DAGs that run once daily). That's part of why I figured the "Workflow Executions" view might be useful, because it shows a row per each log from a workflow execution (which so far seems to involve two log messages, one when the sync starts and has status `ACTIVE` and one when it completes with a status of `FAILED` or `SUCCEEDED`).

I'll plan to walk the team through the dbt in Eng Weekly. It's not the easiest setup to dive into without some guidance!

PR #207, opened 2023-08-04
DE 440 jazzhr dbt tests

kanelouise:
Part 1 of [DE-440](https://techallies.atlassian.net/browse/DE-440)

There are 4 tables that don't have great primary keys outside of `airbyte` added columns. Those columns are: hires, applicants, applicants_to_jobs, categories_to_applicants. All other tables have unique tests on `id`. 

All tests pass!

[DE-360]: https://techallies.atlassian.net/browse/DE-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DE-440]: https://techallies.atlassian.net/browse/DE-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt-cta/jazzhr/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/jazzhr/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -51,6 +51,9 @@ models:
               column_type: timestamp
       - name: _airbyte_tasks_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
 
   - name: hires_base
@@ -89,6 +92,9 @@ models:
               column_type: timestamp
       - name: _airbyte_hires_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: categories_base
     description: ''
@@ -123,6 +129,9 @@ models:
               column_type: timestamp
       - name: _airbyte_categories_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: questionnaire_answers_base
     description: ''
@@ -176,6 +185,9 @@ models:
               column_type: timestamp
       - name: _airbyte_questionnaire_answers_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: applicants_base
     description: ''
@@ -213,6 +225,9 @@ models:
               column_type: timestamp
       - name: _airbyte_applicants_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: applicants_to_jobs_base
     description: ''
@@ -248,6 +263,9 @@ models:
               column_type: timestamp
       - name: _airbyte_applicants_to_jobs_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: jobs_base
     description: ''
@@ -312,6 +330,9 @@ models:
               column_type: timestamp
       - name: _airbyte_jobs_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: activities_base
     description: ''
@@ -352,6 +373,9 @@ models:
               column_type: timestamp
       - name: _airbyte_activities_hashid
         description: ''
+        tests:
+          - not_null
+          - unique
 
   - name: categories_to_applicants_base
     description: ''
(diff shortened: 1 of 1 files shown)


dbt-cta/jazzhr/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,446 @@
...
+        description: ''
+        tests:
+          - dbt_expectations.expect_column_values_to_be_of_type:
+              column_type: timestamp
+      - name: _airbyte_normalized_at
+        description: ''
+        tests:
+          - dbt_expectations.expect_column_values_to_be_of_type:
+              column_type: timestamp
+      - name: _airbyte_tasks_hashid
```
jitoquinto, line 52: Mostly just curious, do we add any unique/not null tests for hash ids??

jitoquinto (dismissed):
lgtm! just left a question about hash ids

dbt-cta/jazzhr/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
emily-flambe: I think we should indeed add unique+not-null tests for all hashid fields. By definition they should be unique (based on how they are constructed in `*_ab3`, and it would just be plain weird if they were ever null lol.

Currently we don't have any cute automated way to do that (though one can do a clever find-and-replace to add them all at once), but there is actually a ticket in the Priority Backlog for making that one of the things that happens by default when we generate dbt tests - [DE-444](https://techallies.atlassian.net/browse/DE-444)!

@kanelouise would you mind adding not-null and unique to all the hashid fields? 🙏  Once those are in then this has my official sign-off!

kanelouise: good q, the hash id's weren't always unique.... specifically for the tables: `applicants`, `applicants_to_jobs`, `categories_to_applicant`, and `questionaire_answers`. 

This is why I added the unique and not null tests to `_airbyte_ab_id` instead.

emily-flambe: Ah!! Ok, I see what happened.

The partition-replacement logic only works when the sync is run consistently. If there's a gap between runs (as happens between phases 2 and 3), you end up with `*_base` tables that have duplicate rows, which contain the exact same data but with `_airbyte_emitted_at` values that are more than a couple of days apart.

I ran through the following exercise for `applicants_to_jobs`, and you can do the same for those other tables to see what I'm talking about.

This first query shows you that the rows are indeed duplicates, and you can see the `_airbyte_emitted_at` values are a few weeks apart (hence the date partition replacement strat will fail to remove the old partition... it's too old 🫠)

```with dupe_hashes as (
select _airbyte_applicants_hashid, count(1)
FROM `prod8b61f23e.partner_a_jazzhr.applicants_base`
group by 1
having count(1)>1
)
SELECT * FROM `prod8b61f23e.partner_a_jazzhr.applicants_base` as table
JOIN dupe_hashes using(_airbyte_applicants_hashid)
order by _airbyte_applicants_hashid```

If you manually delete the rows with the old `_airbyte_emitted_at` values, the above query should return zero rows, and you should see `unique` tests pass for the hashids.

@kanelouise let me know if that makes sense and if you're able to replicate that for the other tables. In principle we should consider it a requirement that `_hashid` fields are unique, since that is more-or-less the reason they exist - they are calculated by hashing over the columns that should define unique rows in a dataset (providing a DIY primary key in case one does not already exist).

And.... is this scuffed and confusing? Yep 👼 open to ideas about how to make this less scuffed and confusing! though we will all internalize this quirk if we run into this issue enough lol

kanelouise: ahh ok, that makes sense! i thought it had to do with some of the `id` field nonesense. but you're right--it makes sense that the `hash_id` should always be unique. fixed that and pushing the update. ran it locally and all is good.

emily-flambe (approved):
:shipit: ! 🔥 🔥 🔥

PR #211, opened 2023-08-09
feat: add tests for USVote Foundation sync [DE-221]

emily-flambe:
While working on this, I discovered that some of the raw data ingested from the USVote Foundation API has duplicate rows at the source. To see for yourself:

```
SELECT
JSON_EXTRACT_SCALAR(_airbyte_data,'$.id') as id, count(1)
FROM `PROJECT.DATASET._airbyte_raw_regions`
group by 1 order by 2 desc
```

Then, using one of those ids that have 2 rows (yes this could be a single query with a CTE, I'm a busy woman)

```
SELECT
_airbyte_data
FROM `PROJECT.DATASET._airbyte_raw_regions`
where JSON_EXTRACT_SCALAR(_airbyte_data,'$.id')='VALUE'
```

If you run those queries you will see there are fully duplicated rows in the raw data itself. This doesn't seem to be true for all of the tables, but it seemed easiest to just add a step in `0_ctes` that would dedupe by taking only a single row for each value of `*_hashid`.

After making that change, all of the tests in `1_cta_full_refresh/_1_cta_full_refresh__models.yml` are passing for me when run locally.

### DIFF
--- a/dbt-cta/usvote_foundation/models/0_ctes/offices_ab4.sql
+++ b/dbt-cta/usvote_foundation/models/0_ctes/offices_ab4.sql
@@ -1,0 +1,18 @@
+{{ config(
+    cluster_by = "_airbyte_emitted_at",
+    partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
+    unique_key = '_airbyte_offices_hashid'
+) }}
+
+-- SQL model to dedupe on the hashid field calculated in *_ab3
+-- (this step is included because the data source seems to include duplicate rows)
+-- depends_on: {{ ref('offices_ab3') }}
+
+SELECT * FROM 
+(
+SELECT 
+    *,
+    ROW_NUMBER() OVER (PARTITION BY _airbyte_offices_hashid ORDER BY _airbyte_ab_id desc) as rownum 
+FROM {{ ref('offices_ab3') }}
+)
+where rownum=1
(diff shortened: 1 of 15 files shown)


royconst:
Just started looking and Huh. Never seen OVER or PARTITION before. Looking them up but guessing that's where you're deduping.

dbt-cta/usvote_foundation/models/0_ctes/offices_ab4.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- SQL model to dedupe on the hashid field calculated in *_ab3
+-- (this step is included because the data source seems to include duplicate rows)
+-- depends_on: {{ ref('offices_ab3') }}
+
+SELECT * FROM 
+(
+SELECT 
+    *,
+    ROW_NUMBER() OVER (PARTITION BY _airbyte_offices_hashid ORDER BY _airbyte_ab_id desc) as rownum
```
jitoquinto, line 15: would it be better to order by `_airbyte_emitted_at`? that way were always grabbing the latest record

emily-flambe: oh, yeah good call! since these are dupes in the actual raw data I think the difference between `_airbyte_emitted_at` values would be on the order of seconds or even milliseconds, but as long as we're ordering by something, it's slightly less arbitrary lol. I'll push that change!

emily-flambe:
> Just started looking and Huh. Never seen OVER or PARTITION before. Looking them up but guessing that's where you're deduping.

yep, welcome to window functions! They're a pretty important SQL concept and definitely worth a small journey of discovery to wrap your head around them, because I guarantee they'll be relevant to you at some point 🌈

jitoquinto (approved):
🚀

royconst (approved):
looks good :thumbsup:

PR #212, opened 2023-08-10
De 312 quickbooks tot total transformation

royconst:
Getting an error running `dbt run --target cta --select tag:cta;`

```
Compilation Error in model vendors_MetaData (tot_quickbooks/models/2_partner_matviews/vendors_MetaData.sql)
  Model 'model.default.vendors_MetaData' (tot_quickbooks/models/2_partner_matviews/vendors_MetaData.sql) depends on a source named 'cta.vendors_MetaData_base' which was not found
```


dbt-cta/tot_quickbooks/models/sources.yml
```diff
@@ -0,0 +1,315 @@
+version: 2
+sources:
+- name: tot_quickbooks_schema
```
kanelouise, line 3: this should be `name: cta`

royconst:
Changed the sync/folder name. Both cta and partner dbt targets are running successfully.

royconst:
This is my first phase 2 ticket so please get nit-picky

emily-flambe:
@royconst oh man, what a heftychomnk of a ticket!! I don't see any flags in the code, but here are a couple more steps you can take to kick the tires on this whole deal:

1) try running the `dbt run --target cta --select tag:cta` command locally using the data in prod, to make sure it cooperates with the full data (since it looks like in dev we are only using the CTA test account and syncing a single row)
2) add the `dev_deploy` label to both this PR and your PR in airflow-dags to get this running in dev Composer!

dbt-cta/quickbooks/models/0_ctes/bills_Line_ItemBasedExpenseLineDetail_ab2.sql
```diff
@@ -0,0 +1,21 @@
+{{ config(
+    cluster_by = "_airbyte_emitted_at",
+    partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
+    unique_key = "_airbyte_ab_id"
+) }}
+-- SQL model to cast each column to its adequate SQL type converted from the JSON schema type
+-- depends_on: {{ ref('bills_Line_ItemBasedExpenseLineDetail_ab1') }}
+select
+    _airbyte_Line_hashid,
+    cast(UnitPrice as {{ dbt_utils.type_float() }}) as UnitPrice,
```
royconst, line 10: This field was failing with the autogenerated airbyte dbt. It was originally set as an int. It was trying to cast a number like `30.5` and throwing an error. It definitely seems like it is always a currency. I changed every instance of `UnitPrice` to a float as it seems like it's a currency for all of them. We were just getting one error because the other instances weren't populated in our pulled data.

royconst:
> @royconst oh man, what a heftychomnk of a ticket!! I don't see any flags in the code, but here are a couple more steps you can take to kick the tires on this whole deal:
> 
> 1. try running the `dbt run --target cta --select tag:cta` command locally using the data in prod, to make sure it cooperates with the full data (since it looks like in dev we are only using the CTA test account and syncing a single row)
> 2. add the `dev_deploy` label to both this PR and your PR in airflow-dags to get this running in dev Composer!

Re: # 1 It did run `--target` cta and partner successfully on prod. The latter to partner-a and I could see the views.

Re: # 2 Yea, I was wondering if we wanted to keep that dev setup but it would probably be good. I wonder how it will run with no data actually syncing. Hopefully/probably fine. Will test now.

royconst:
I setup dev. Running it locally and `--target cta` ran just fine. I can see the new tables. Running `--target partner` had an issue though. It hung on a materialized view. When cancelling I got `The bigquery adapter does not support query cancellation. Some queries may still be running!`. Thinking it's nothing crazy and a request timed out or something. Running the partner views again with `--full-refresh` and it's progressing.

royconst:
`--target partner` ran fine the second time on dev :thumbsup:

jitoquinto:
@royconst oh wow this connector has a ton of models! The dbt looks good to me, the one comment i would have is that Airbyte default normalization is pretty trash sometimes because it breaks out any nested field into its own table. which doesnt always make sense (like instances where its a single column? what are you doing Airbyte 🤷 ). So sometimes i end up just deleting a lot of those tables created from nested fields and just extract the JSON directly into the main stream table. It definitely helps cut down on a lot of tables that probably wont be used. Probably not something for this connector right now considering just how many models there are but figured id note it! oh and @emily-flambe did start making a version of default dbt normalization that we can use internally! so hopefully we wont even have to worry about it in the near future haha

jitoquinto (approved):

dbt-cta/quickbooks/models/0_ctes/bills_Line_ItemBasedExpenseLineDetail_ab2.sql
emily-flambe: This sounds like the correct approach to me! Gotta love how Airbyte's default dbt is like at least 99% fine (well except for the nested fields like Jesus mentioned lol) except for when it's not 🫠

emily-flambe (approved):
Thanks for all the extra science @royconst !! This dbt looks good to go, my one request (not a blocker for this PR) would be to configure an Airbyte sync in dev using the actual (non-test) credentials, so that we have the test sync running with legit data. This PR is good to merge though!

kanelouise (approved):
PHEW, as the other BANDDE's have stated: bad ass work on a chonker of a Phase 2! LGTM.

royconst:
@emily-flambe totally understand the intent behind using real data but, minor gripe, that would require another live meeting with the partner to get credentials, I've bothered them enough with that lol. Mainly though, it's very sensitive content that I wouldn't feel comfortable putting it in an environment where we do testing. This dev setup really won't tell us a whole lot, but at least it will notify if the api is down, changes in schema, or if we need to test upgrading the connector.

emily-flambe:
> @emily-flambe totally understand the intent behind using real data but, minor gripe, that would require another live meeting with the partner to get credentials, I've bothered them enough with that lol. Mainly though, it's very sensitive content that I wouldn't feel comfortable putting it in an environment where we do testing. This dev setup really won't tell us a whole lot, but at least it will notify if the api is down, changes in schema, or if we need to test upgrading the connector.

Ah!!  I had forgot that the additional connection would require another meeting with the partner. So nuts to that, let's keep running the CTA test connection in dev. Thanks for explaining!!

PR #213, opened 2023-08-17
feat (spoke_mo): use incremental logic to update base tables

emily-flambe:
`dbt run` and `dbt test` work locally for me using the prod data with the failing tests (after deleting the dupe from `_airbyte_raw_message` and `message_base`, see Slack).


dbt-cta/spoke_mo/models/1_cta_incremental/zip_code_base.sql
jitoquinto, line 1: looks like this one needs the partitions removed at the top!

jitoquinto (commented):
looks like the last model needs to have the partitions removed but all else lgtm!

dbt-cta/spoke_mo/models/1_cta_incremental/zip_code_base.sql
emily-flambe: ope actually zip_code is the one stream in this sync that IS full refresh overwrite, so it gets to go be alone in a 1_cta_full_refresh folder! Gonna make that change presently

emily-flambe: ok what do we think about this commit? https://github.com/community-tech-alliance/dbt-cta/pull/213/commits/3ce43b421acef1ef3afb40f512cbd1d962912a45

just ran these commands locally and they succeeded: 

```
pipenv run dbt run --target cta --select 1_cta_full_refresh
pipenv run dbt test --target cta --select 1_cta_full_refresh
```

kanelouise (approved):
LGTM, this special snowflake i think was the reason we stuffed them all in the same folder at first.

dbt-cta/spoke_mo/models/1_cta_incremental/zip_code_base.sql
jitoquinto: nice! :shipit:

jitoquinto (approved):

emily-flambe:
> LGTM, this special snowflake i think was the reason we stuffed them all in the same folder at first.

yeah lol it's definitely unusual to have a sync with ONE stream that's full refresh while the others aren't. Computers 🫠

PR #214, opened 2023-08-18
feat: Add sqlfluff config and test out on a project

jitoquinto:
Alright i tried out both `sqlfmt` and `sqlfluff` and using the `sqlfluff fix` command seems like a better approach for formatting. It is slower than `sqlfmt` but you can specify which rules you want ignored and just overall more customizable. I excluded 2 rules from sqlfluff config

- CP02: `Inconsistent capitalisation of unquoted identifiers`
  - Basically, this rule tries to make sure your column names have consistent case (camel case, snake, pascal, all caps, etc..). I let it off because i don't think we should be messing with the column names of existing data at this point. It would involve doing a reset of all tables and would probably cause problems to partners who have stuff built on the data. But let me know if ya think we should do this and we can try to figure out a plan for it!
- Max Line Length
  - I just turned this off because i thought it made the files look uglier haha. The unnesting dbt files get very long so they were getting split into multiple lines and i thought it was harder to read that way.


emily-flambe:
I haven't yet looked at this deeply but just want to say I agree with your aesthetic judgment with respect to max line length (I think I have seen parts of our dbt code run through a SQL linter that respected that limit and it was hideous) - and also yeah if we pivot on the capitalization stuff now we will probably break things. I think the time to set standards about that was... probably about 16 months ago lol

jitoquinto:
> I haven't yet looked at this deeply but just want to say I agree with your aesthetic judgment with respect to max line length (I think I have seen parts of our dbt code run through a SQL linter that respected that limit and it was hideous) - and also yeah if we pivot on the capitalization stuff now we will probably break things. I think the time to set standards about that was... probably about 16 months ago lol

lol yeahhh we could do something like for any new stuff were adding, run with that rule enabled. I guess that would just kinda have to live in our brains and be something we do though? or i guess i could add it into the dbt_helper script and run through the linter when ever its creating a new dbt project?

emily-flambe:
> > I haven't yet looked at this deeply but just want to say I agree with your aesthetic judgment with respect to max line length (I think I have seen parts of our dbt code run through a SQL linter that respected that limit and it was hideous) - and also yeah if we pivot on the capitalization stuff now we will probably break things. I think the time to set standards about that was... probably about 16 months ago lol
> 
> lol yeahhh we could do something like for any new stuff were adding, run with that rule enabled. I guess that would just kinda have to live in our brains and be something we do though? or i guess i could add it into the dbt_helper script and run through the linter when ever its creating a new dbt project?

Oh I do like the idea of doing some linting in the dbt_helper script! That sounds like the simplest strat for applying rules about capitalization only to new dbt projects without messing up anything we already have.

emily-flambe:
OK I just took a closer look at the SQL files in this PR that were linted, and I like it! Would definitely be a significant style cleanup to apply this to all our existing projects. Real professional-like!

@jitoquinto what would be involved in setting this up in a pre-commit hook? Back in the day (possibly even before you were with CTA??) we used Husky for pre-commit hooks... for what I'm no longer even sure (maybe terraform?). I think husky was OK, just required some setup, but probably all things would require setup... curious if there are any tools or strats you've used in the past 🤔

jitoquinto:
> @jitoquinto what would be involved in setting this up in a pre-commit hook? Back in the day (possibly even before you were with CTA??) we used Husky for pre-commit hooks... for what I'm no longer even sure (maybe terraform?). I think husky was OK, just required some setup, but probably all things would require setup... curious if there are any tools or strats you've used in the past 🤔

@emily-flambe i've only ever used [pre-commit](https://pre-commit.com/) which uses git-hooks to run whenever you do a git commit. Here is an example of what I'm doing now. It just runs the linter which will tell you if there are any failures in linting and not allow you to commit until they are fixed. 
image

So the you would need to manually run the `sqlfluff fix` (or the helper_script) to auto-format similar to black

emily-flambe:
> > @jitoquinto what would be involved in setting this up in a pre-commit hook? Back in the day (possibly even before you were with CTA??) we used Husky for pre-commit hooks... for what I'm no longer even sure (maybe terraform?). I think husky was OK, just required some setup, but probably all things would require setup... curious if there are any tools or strats you've used in the past 🤔
> 
> @emily-flambe i've only ever used [pre-commit](https://pre-commit.com/) which uses git-hooks to run whenever you do a git commit. The config file for pre-commit is just a yaml file too so its pretty easy to work with. Here is an example of what I'm doing now. It just runs the linter which will tell you if there are any failures in linting and not allow you to commit until they are fixed. image
> 
> So the you would need to manually run the `sqlfluff fix` (or the helper_script) to auto-format similar to running black.
> 
> edit: oh it also only runs on files that are staged for commit! so thats nice. Im gonna work on the helper script so that also only runs on files that were changed and not the whole dang dbt repo update: i wasnt able to get it to run on just diffed files, at least not the fix option 🙃

Aww yeah pre-commit looks good! I don't mind having to actually run the `sqlfluff fix` command instead of the pre-commit hook doing it automatically (though if we do want that, I think that's what husky does). We already have to manually `black` things after all, and I mean it's not fully automated luxury space communism but it's fine 😛

emily-flambe (dismissed):
Code LGTM!!! possibly we could add the note about committing with `--no-verify` to the README but it's not urgent because that's pretty easily googleable 😉 

I am VERY excited for our fluffy new lives

jitoquinto:
> Code LGTM!!! possibly we could add the note about committing with `--no-verify` to the README but it's not urgent because that's pretty easily googleable 😉
> 
> I am VERY excited for our fluffy new lives

ooh yeah good point, adding it in there :cat-typing:

PR #215, opened 2023-08-21
style: format action builder

jitoquinto:
Ran SQLFluff on Action Builder! The only changes im seeing on here are 
- Remove unused table aliases
- Add `as` keyword when declaring an alias
- remove empty lines at end of file

There were some other linting errors that were all of this kind:
``` Keywords should not be used as identifiers.```
But they were column existing column names that we probably shouldn't change, so i left it as is.


kanelouise (commented):
seems cool! i have the same q here as the other `sqlfluff` PR you have open right now: does dbt run without a problem without the extra line at the end of each file?

jitoquinto:
> seems cool! i have the same q here as the other `sqlfluff` PR you have open right now: does dbt run without a problem without the extra line at the end of each file?

same here! ran the models and it ran through successfully
image

kanelouise (approved):
nice! glad the dbt runs!

emily-flambe (approved):
FLUFF ALL THE THINGS. And yeah sqlfluff is gonna have to accept that we have keywords as column names sometimes and that's just the world we live in

PR #216, opened 2023-08-21
style: run sqlfluff on action_network models

jitoquinto:
Ran SQLFluff on Action Network. The only changes im seeing on here are 
- Remove unused table aliases
- Add `as` keyword when declaring an alias
- remove empty lines at end of file

There were some other linting errors that were all of this kind:
``` Keywords should not be used as identifiers.```
But they were column existing column names that we probably shouldn't change, so i left it as is.


kanelouise (commented):
cool! does dbt still run ok without the empty line at the end of the files?

jitoquinto:
> cool! does dbt still run ok without the empty line at the end of the files?

@kanelouise good point! I ran the dbt locally and it ran through successfully so im guessing the line or any of the other changes dont affect it!

image

kanelouise (approved):
awesome! lgtm then :)

PR #217, opened 2023-08-22
style: format adp_worker_management models

jitoquinto:
Ran SQL Fluff on adp_worker_management models and tested dbt for both CTA and Partner in dev


emily-flambe (approved):
I see the fluff disagrees with my aesthetic choices about capitalized function words. VERY WELL

jitoquinto:
> I see the fluff disagrees with my aesthetic choices about capitalized function words. VERY WELL

lol we can change it to be caps if we want to! (you can see it in the `.sqlfluff` file) it just defaults to this but im not entirely sure how dbt macros work 🤔 like do they need to be the same capitalization as they're definition?

jitoquinto:
> haha I have no actual emotional attachment to the capitalized words! I say let's just keep going with how we've been fluffing things so far 😎

:cat-jam:

PR #218, opened 2023-08-23
[DE-467] feat: create dbt project for `bigquery_monitoring`

emily-flambe:
This will enable us to generate authorized views in partner projects that give them access to reporting views with handy information about queries run in their project. (See accompanying PR in airflow-dags that adds this project to the DMDF config [here](https://github.com/community-tech-alliance/airflow-dags/pull/402))

### DIFF
--- a/dbt-cta/bigquery_monitoring/models/2_reporting/job_details.sql
+++ b/dbt-cta/bigquery_monitoring/models/2_reporting/job_details.sql
@@ -1,0 +1,6 @@
+SELECT jobs.*
+FROM {{ ref('json_extract_job_details') }} as jobs
+JOIN {{ ref('distinct_job_names') }} as distinct_jobs using(jobName)
+where
+  jobs.resource_type = 'bigquery_project'
+  and array_length(jobs.referencedTables)>0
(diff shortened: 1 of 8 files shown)


dbt-cta/bigquery_monitoring/models/2_reporting/job_details.sql
```diff
@@ -0,0 +1,6 @@
+SELECT jobs.*
+FROM {{ ref('json_extract_job_details') }} as jobs
+JOIN {{ ref('distinct_job_names') }} as distinct_jobs using(jobName)
+where
+  jobs.resource_type = 'bigquery_project'
+  and array_length(jobs.referencedTables)>0
```
jitoquinto, line 6: What is the purpose for this array length check??

dbt-cta/dbt_project.yml
```diff
@@ -62,11 +62,26 @@ models:
 
     # for the /monitoring project
     0_staging:
+      +tags:
```
jitoquinto, line 65: Are these being used by other models? Like this wont accidentally grant access to some tables we dont want granted access to?

jitoquinto (commented):
Oooh this is really cool! So will this create a new dataset in the partners project with these views?

dbt-cta/dbt_project.yml
jitoquinto: sweet thanks for the helpful commentary!

jitoquinto (approved):
wowow thanks for demoing too!

PR #219, opened 2023-08-23
yeet: remove `monitoring` dbt project files

emily-flambe:
We are moving this dbt to our private CTA repo for internal reporting. Thus from the perspective of this repo, the dbt is no longer - gone but not forgotten


kanelouise (approved):
:cat-salute:

PR #227, opened 2023-08-25
feat (bq monitoring): filter out logs for queries run by CTA staff

emily-flambe:
@michaelefisher if we want to exclude these rows from the views entirely, this is how we'd do it 👀


royconst (approved):
Looks fine!

PR #228, opened 2023-09-06
[DE-480] feat: bump elementary dbt package

jitoquinto:
@emily-flambe I forgot that I had done this in Dev already! but we will need to make this change in prod and update the dbt-package for elementary in prod GCS if we merge in the other PR bumping the python package


emily-flambe:
ope, thanks for the tip @jitoquinto ! I know you're going OOO so I can take this from here!!

emily-flambe (approved):

PR #237, opened 2023-09-15
fix (usvotefoundation): fix upstream model references in models for nested data

emily-flambe:
`dbt run` is succeeding for me locally after making this fix. But how did the dbt ever succeed in the first place...?

### DIFF
--- a/dbt-cta/usvote_foundation/models/0_ctes/offices_addresses_ab1.sql
+++ b/dbt-cta/usvote_foundation/models/0_ctes/offices_addresses_ab1.sql
@@ -4,8 +4,8 @@
     tags = [ "nested-intermediate" ]
 ) }}
 -- SQL model to parse JSON blob stored in a single column and extract into separated field columns as described by the JSON Schema
--- depends_on: {{ ref('offices') }}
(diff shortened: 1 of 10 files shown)


jitoquinto:
> `dbt run` is succeeding for me locally after making this fix. But how did the dbt ever succeed in the first place...?

do you think maybe the `json_extract_array` function isnt working how we expect it to anymore? like maybe it was creating an [array before here](https://github.com/community-tech-alliance/dbt-cta/blob/main/dbt-cta/usvote_foundation/models/0_ctes/offices_ab1.sql#L19) but now its just making it json?

emily-flambe:
> left a comment, but maybe the functionality of one of the extract json macros changed?? i have no other theories as to why this just came up now 🙃

Well, I have no other theories lol, but you'd think if a macro like that had changed then we'd see mysterious UNNEST-related errors in a bunch of our syncs, because we unnest arrays somewhat frequently I think 🤔 🤔 🤔 🤔  I'm emotionally willing to keep this spooky mystery somewhat-unsolved until/unless we see the same thing start to happen again...

PR #239, opened 2023-09-26
hotfix: regenerate schema yamls with new dbt version (unclear why this helped)

emily-flambe:
These yamls are already saved to GCS and the blocks DAG is proceeding successfully, so I won't merge this until BANDDE review (cc @jitoquinto @kanelouise @royconst )

### DIFF
--- a/dbt-cta/blocks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/blocks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -135,7 +135,6 @@ models:
         description: ''
       - name: _airbyte_normalized_at
         description: ''
-        tests:
       - name: _airbyte_registration_forms_hashid
         description: ''
         tests:
@@ -727,6 +726,32 @@ models:
         tests:
           - unique
 
+  - name: admin_users_base
+    description: ''
+    columns:
+      - name: session
+        description: ''
+      - name: id
+        description: ''
+        tests:
+          - unique
+      - name: email
+        description: ''
+      - name: encrypted_password
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+        tests:
+          - unique
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_admin_users_hashid
+        description: ''
+        tests:
+          - unique
+
   - name: state_party_codes_base
     description: ''
     columns:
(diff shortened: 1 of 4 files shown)


dbt-cta/blocks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -727,6 +726,32 @@ models:
...
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+        tests:
+          - unique
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_admin_users_hashid
```
jitoquinto, line 750: Should these have any non-null checks?

jitoquinto (approved):
This all looks good to me, i did notice there weren't any not null tests for id's. Is that something we want in there or nah?

dbt-cta/blocks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
emily-flambe: we removed them because composer was yeeting the tasks because there were too many tests and they were taking too long, RIP

kanelouise (approved):

emily-flambe:
> This all looks good to me, i did notice there weren't any not null tests for id's. Is that something we want in there or nah?

This DAG was failing a few weeks ago because there were So Many Tests that the Airflow tasks were choking - that's why `not_null` has been removed, along with the type checks for timestamps, since really the main thing we care about is uniqueness.

Come to think of it, it's actually not clear to me whether there's ever a benefit to having both `unique` and `not_null` - if multiple rows are null for that field, they wouldn't be unique, would they...? dbt docs show both tests on their page about tests ([here](https://docs.getdbt.com/docs/build/tests)) but I'm skeptical 🧐

jitoquinto:
> > This all looks good to me, i did notice there weren't any not null tests for id's. Is that something we want in there or nah?
> 
> This DAG was failing a few weeks ago because there were So Many Tests that the Airflow tasks were choking - that's why `not_null` has been removed, along with the type checks for timestamps, since really the main thing we care about is uniqueness.
> 
> Come to think of it, it's actually not clear to me whether there's ever a benefit to having both `unique` and `not_null` - if multiple rows are null for that field, they wouldn't be unique, would they...? dbt docs show both tests on their page about tests ([here](https://docs.getdbt.com/docs/build/tests)) but I'm skeptical 🧐

ohhh yeah! :thinking: yeah thats a good point maybe they aren't doing anything extra haha

PR #240, opened 2023-09-27
clean(action_builder): remove tests of `_airbyte_*` fields (deprecated)

emily-flambe:
CTA now syncs Action Builder using an Airflow DAG, so in order to run dbt tests on the data using the code in this repository, we need to clear out any tests on columns prefixed `_airbyte_*`, as these no longer exist in the base tables created by the DAG.


jitoquinto (approved):
👋

PR #252, opened 2023-10-24
add impactive models and fix dbt extractor

michaelefisher:
Here's the dbt after running:
```
  Format exported Airbyte dbt to CTA structure
```
and choosing `impactive` and `utils/config/` as the location of the extracted dbt

### DIFF
--- a/dbt-cta/impactive/models/0_ctes/activities_export_ab1.sql
+++ b/dbt-cta/impactive/models/0_ctes/activities_export_ab1.sql
@@ -1,0 +1,37 @@
+{{ config(
+    cluster_by = "_airbyte_emitted_at",
+    partition_by = {"field": "_airbyte_emitted_at", "data_type": "timestamp", "granularity": "day"},
+    unique_key = "_airbyte_ab_id"
+) }}
+-- SQL model to parse JSON blob stored in a single column and extract into separated field columns as described by the JSON Schema
+-- depends_on: {{ source('cta', '_airbyte_raw_activities_export') }}
+select
+    {{ json_extract_scalar('_airbyte_data', ['performs'], ['performs']) }} as performs,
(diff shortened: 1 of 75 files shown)


emily-flambe:
> and choosing `impactive` and `utils/config/` as the location of the extracted dbt

Is this operating as intended? Looks like the new Impactive dbt files ended up in dbt-cta/impactive/models (which is where they should ultimately end up) - did you move the files there by hand?

emily-flambe (changes requested):
Can we please modify the script so that the filenames for the CTA base models (in `1_cta_*`) have the `_base` suffix?

emily-flambe:
> Can we please modify the script so that the filenames for the CTA base models (in `1_cta_*`) have the `_base` suffix?

Ope, I see now that this is a separate option one can select when firing up the script, `APPEND_BASE_TO_FILES`. Not a big deal, would be cute to automatically run that after extracting the dbt, but not a blocker to merging this in... thus I approve

emily-flambe (approved):
Can we please modify the script so that the filenames for the CTA base models (in `1_cta_*`) have the `_base` suffix?

^^ would be cute but not actually a blocker since there's a separate option for performing this step in the script

michaelefisher:
@emily-flambe I didn't do anything by hand, I just chose the options that Jesus had already put together. This was ultimately just subbing out the actual extraction piece.

I don't think that changing the options for the script are in-scope here, I wasn't intending to touch any of the other script features that already exist. Just getting it to parity.

michaelefisher:
Lol ok @emily-flambe I can see why you were asking about where the files went. This was deleting them:

https://github.com/community-tech-alliance/dbt-cta/pull/252/files#diff-7107c0b9edf473b2867bdfdb196193ccdf2d9644c28d35acacc554bfc9409c81L19

Whoops.

emily-flambe (approved):

PR #253, opened 2023-10-31
feat (Quickbooks): add dbt tests [DE-530]

emily-flambe:
This PR adds uniqueness tests for all `*_hashid` fields (when they are unique per row) and some `id` fields. (Not all tables with `id` use that field as a PK 😐 ). I left out other tests we often include because of the sheer number of tables present - Airflow has a tendency to fail when you give it too many things to do in a single dbt task.

These tests are succeeding when run locally using prod data.

cc @kanelouise @royconst @jitoquinto

### DIFF
--- a/dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -6871,8 +6871,6 @@ models:
         description: ''
       - name: id
         description: ''
-        tests:
-          - unique
       - name: city
         description: ''
       - name: line1
--- a/dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
+++ b/dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
@@ -1,0 +1,6920 @@
+version: 2
+
+models:
+
+  - name: tax_agencies_MetaData_base
+    description: ''
+    columns:
+      - name: _airbyte_tax_agencies_hashid
+        description: ''
+      - name: createtime
+        description: ''
+      - name: lastupdatedtime
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_MetaData_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: items_IncomeAccountRef_base
+    description: ''
+    columns:
+      - name: _airbyte_items_hashid
+        description: ''
+      - name: name
+        description: ''
+      - name: value
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_IncomeAccountRef_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: estimates_TxnTaxDetail_TaxLine_TaxLineDetail_base
+    description: ''
+    columns:
+      - name: _airbyte_TaxLine_hashid
+        description: ''
+      - name: percentbased
+        description: ''
+      - name: taxrateref
+        description: ''
+      - name: netamounttaxable
+        description: ''
+      - name: taxpercent
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_TaxLineDetail_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: journal_entries_CurrencyRef_base
+    description: ''
+    columns:
+      - name: _airbyte_journal_entries_hashid
+        description: ''
+      - name: name
+        description: ''
+      - name: value
+        description: ''
+      - name: _airbyte_ab_id
+        description: ''
+      - name: _airbyte_emitted_at
+        description: ''
+      - name: _airbyte_normalized_at
+        description: ''
+      - name: _airbyte_CurrencyRef_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: purchase_orders_CurrencyRef_base
+    description: ''
+    columns:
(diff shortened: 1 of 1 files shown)


dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,6918 @@
+version: 2
+
+models:
+
+  - name: tax_agencies_MetaData_base
+    description: ''
+    columns:
+      - name: _airbyte_tax_agencies_hashid
+        description: ''
```
royconst, line 9: Looks like this should probably be unique. [API docs](https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/taxagency) say it's just create/update time so tax agencies wouldn't have multiple.

Since we had a lot of questions about data credibility I'm checking out the api docs just to be safe. Though it might get time consuming.

dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,6918 @@
...
+      - name: _airbyte_MetaData_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: items_IncomeAccountRef_base
+    description: ''
+    columns:
+      - name: _airbyte_items_hashid
+        description: ''
```
royconst, line 29: Think this is also 1-1 items to income account ref. [docs](https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/item)

dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,6918 @@
...
+      - name: _airbyte_TaxLineDetail_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: journal_entries_CurrencyRef_base
+    description: ''
+    columns:
+      - name: _airbyte_journal_entries_hashid
+        description: ''
```
royconst, line 73: Singular so looks to be unique. https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/journalentry

dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,6918 @@
...
+      - name: _airbyte_CurrencyRef_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: purchase_orders_CurrencyRef_base
+    description: ''
+    columns:
+      - name: _airbyte_purchase_orders_hashid
+        description: ''
```
royconst, line 93: Same. 1-1 https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/purchaseorder

dbt-cta/quickbooks/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,6918 @@
...
+      - name: _airbyte_CustomerRef_hashid
+        description: ''
+        tests:
+          - unique
+
+  - name: bills_MetaData_base
+    description: ''
+    columns:
+      - name: _airbyte_bills_hashid
+        description: ''
```
royconst, line 179: same unique. https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/bill

kanelouise (approved):
LGTM, looks like Roy has some insights on primary keys that are possible on some tables.

jitoquinto (approved):
lgtm!

royconst:
@jitoquinto 
> some of these are just broken out tables that Airbyte could have probably just expanded into its own columns instead of creating a whole new dang table for?

Yes and no lol. It could reduce a lot of the tables I think because it's references to other objects in quickbooks. The problem is that the data comes from the API so technically every single object is stateless. I mean I'm assuming true stateless, but at the very least mostly stateless because it's API. 

> maybe changing the dbt to break out instead of creating a new table may be better?

Unfortunately if we did that then the related columns would be a repeatable string that is actually a json object with a bunch of extra signifiers in there like weird definitions of what each field is that doesn't really help and muddles the data up. A lot of them are also arrays because of multiple accounts being tied to one thing etc.

>I'd also be okay with not having uniqueness tests on those hash ids (that relate back to parent tables) since they're getting built out from the parent table which should already make it unique since its 1-1 and full-refresh?

Pulling in @emily-flambe here. Normally yea, I'd 100% say going through everything would take way too much time and we can assume stuff comes in correctly because of full refresh. The only reason I think it might be required is because of the 3rd codeblock on [this comment](https://techallies.atlassian.net/browse/PAD-2064?focusedCommentId=21599) where there are `purchase_Line` objects with `purchase_hashid`'s that do not exist in `purchase`. Also the connector is technically still in alpha and not beta yet.
That said, It would definitely take time to hunt down all the connections through the api docs. I stopped after only a couple hah. IDK if we can say that we really have full dbt tests if we don't though.

royconst:
I think If we decided to make even a bare minimum of "relational" tests it would probably bump the story up a point or two. But we're also working back from originally having duplicates and stuff because we wrote 0 tests in the first delivery....

emily-flambe:
@royconst appreciate you digging in here! time is very precious these days so I don't want to creep the scope on the QB tests unless absolutely necessary. I will add in the additional tests you identified for unique keys and then hope to wrap this up. If we need to do more sophistimacated dbt on these models in the future then we can plan that work separately. Does that sound ok to you?

royconst:
Yea, that works. It would definitely address the duplicates part of [PAD-2064](https://techallies.atlassian.net/browse/PAD-2064). If we want further confirmation on the data quality past dupes we could add steps in there or the missing data story.

emily-flambe:
> Yea, that works. It would definitely address the duplicates part of [PAD-2064](https://techallies.atlassian.net/browse/PAD-2064). If we want further confirmation on the data quality past dupes we could add steps in there or the missing data story.

I believe having uniqueness checks on all the unique `_hashid` fields will be sufficient to prevent duplicates like the ones we resolved earlier this week! as long as every table has a `unique` test that can prevent dupes, I think we should be good.

royconst (approved):
Forgot to actually approve with my last comment

PR #254, opened 2023-11-01
Impactive dbt

kanelouise:
for DE-510: dbt and tests. The tests passed locally in dev + prod.


royconst (approved):
I see no issues but not very versed on phase 2 structure still.

kanelouise:
> I see no issues but not very versed on phase 2 structure still.

Thanks, @royconst! i just added tests, too, so this is ready for final review.

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,1237 @@
...
+      - name: van_database_mode
+        description: ''
+      - name: id
+        description: ''
+        tests:
+          - not_null
+          - unique
+      - name: campaign_id
+        description: ''
+      - name: _airbyte_ab_id
```
jitoquinto, line 37: I'm a bit conflicted with tests on `_airbyte_ab_id` especially if there are other columns that can be used for row uniqueness. I think the name of this column in Destinations V2 is more descriptive (`_airbyte_raw_id`) since its just the id for the record in the raw table. So the test would catch if we somehow parse the same raw record multiple times but in a full refresh model it shouldn't happen? what do you think?

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,1237 @@
...
+        description: ''
+        tests:
+          - dbt_expectations.expect_column_values_to_be_of_type:
+              column_type: timestamp
+      - name: _airbyte_normalized_at
+        description: ''
+        tests:
+          - dbt_expectations.expect_column_values_to_be_of_type:
+              column_type: timestamp
+      - name: _airbyte_teams_export_hashid
```
jitoquinto, line 721: I think hash_id would be a better uniqueness check than airbyte_ab_id if theres no primary key column like id in the other tables

jitoquinto (commented):
left a couple of comments/questions but overall this looks good to me!

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
kanelouise: Hmmm, I'm afraid I don't know the differences between `_airbyte_ab_id` and `_airbyte_raw_id`. I've been striving to use a native primary key when possible in addition to this `_airbyte_ab_id` column and definitely think if there's a better column to use that we should! Like you said in your next comment, pershaps the `_airbyte_{table_name}_hash_id` column might be better?

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
kanelouise: Yeah, there was no obvious pk i could see in this table. Adding the uniqueness check to the hash_id sounds good. Flagging that the `airbyte_ab_id` tests are automagically added from the dbt test script. Would love to chat more with the team, mayhaps, about how we can make the dbt tests a bit tighter?

kanelouise: @jitoquinto - i added a uniqueness test here. would love another peek/approval when you get a chance.

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
jitoquinto: they're the same column! Airbyte just decided to rename the `airbyte_ab_id` in the newest version of destinations. But yes I think the hash_id is a better column for uniqueness check!

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
jitoquinto: yay thank you! And yeah agreed, it would be good to take another look at the script and re-evaluate default ones. I personally don't think were getting much value out of checks on the `_airbyte_ab_id` tests but would be good to discuss! maybe a good eng weekly topic for when @emily-flambe can attend

and yep taking a look :cat_typing:

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -0,0 +1,1240 @@
...
+        description: ''
+      - name: van_id
+        description: ''
+      - name: updated_at
+        description: ''
+      - name: supplied_zip_code
+        description: ''
+      - name: id
+        description: ''
+        tests:
```
jitoquinto, line 278: hmm i think this is missing tests?

kanelouise: `id` is not a primary key, weirdly for this table. it seems to be a `user_id` but the user can be involved in several different campaigns. so for this one, i added a uniqueness test to the hash id.

kanelouise: (i added the same uniqueness test to the similar teams export table as well.)

jitoquinto: ahh okay gotcha! i saw the empty `tests:` and thought it may be missing some but makes sense now thanks!

jitoquinto (approved):
Thanks for answering all my questions! lgtm! :shipit:

dbt-cta/impactive/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
kanelouise: ahh i see what you're saying! i took that empty test field out. one more approval request coming your way.

jitoquinto (approved):

PR #255, opened 2023-11-09
[PAD-2064] hotfix (Quickbooks): append `_base` suffix to refs in CTEs for tables from nested fields

emily-flambe:
This is succeeding locally and given timing considerations I want to make sure this gets merged in, so YOLOing now - cc @royconst

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,13 @@ for file in $(ls mobilize/models/1_cta_base_tables/); do
     mv mobilize/models/1_cta_base_tables/"$file" mobilize/models/1_cta_base_tables/"$newFile"
 done
 ```
+
+## IMPORTANT note about tables generated from nested fields
+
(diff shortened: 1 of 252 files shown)


royconst:
I can type it out real quick but can we add a note to the [README](https://github.com/community-tech-alliance/dbt-cta/tree/main/utils#add-_base-suffix-to-all-models-in-a-folder) so this doesn't happen again?

emily-flambe:
> maybe something like this. Not super clear but better than not having it.
> 
> ```
> Inside `0_ctes` we might need to update the reference generated by airbyte. Since we use `_base` as  a postfix we might need to update these lines.  
> `from {{ ref('accounts') }} as table_alias` would change to -> `from {{ ref('accounts_base') }} as table_alias`
> ```

Ah yeah, we do need to document this - it's an unusual step which is why we weren't on the lookout for it. I'll add a note to the README before merging!

PR #256, opened 2023-11-25
use `jsonpath="{.status.phase}"` to get status of pod

emily-flambe:
@michaelefisher I was trying to use copy_dbt_from_k8s_pod.sh (I have my reasons) and this seems like a more straightforward way to check for whether a pod is in a Running state. (Also the code as currently written doesn't work for me - I can't use the existing jsonpath to access the status of a pod, even when I run the command manually in terminal)

However even after fixing(?) this part, I'm still not able to find any dbt files in the running pods. Here's what happens when I run the shell script:

image

Setting aside the `error from server (BadRequest)`, the list of files the script is attempting to copy doesn't include anything dbt-related. So my next step was to attempt to look around the pod filesystems manually to see if I could find the dbt files. But (predictably) the `config/` folder just looks like this, and I can locate no other files:

image

This was the result for both the source and destination pods, which were both Running at the time:

image

I couldn't find instructions for running this shell script so maybe I'm doing something wrong, anything jump out at you?

This is the Airbyte connection I'm playing with (I'm doing some science for P2P dev, it's nothing relating to current partner sync development): https://k8s.cta-tech.dev/workspaces/4f1821e9-c1a2-4b22-9265-120d2054355f/connections/5df6404b-089d-4cc7-8cca-ba9ea737a190/status

### DIFF
--- a/utils/copy_dbt_from_k8s_pod.sh
+++ b/utils/copy_dbt_from_k8s_pod.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+shopt -s nocasematch
+
 # Ensure the user provides a pod name regex pattern, namespace, and container name as arguments
 if [ "$#" -ne 3 ]; then
     echo "Usage: $0 <POD_NAME_REGEX_PATTERN> <NAMESPACE> <CONTAINER_NAME>"
@@ -18,7 +20,7 @@ while true; do
         # Use grep to filter pods based on the regex pattern
         if [[ $POD =~ $POD_NAME_PATTERN ]]; then
             # Fetch the container status using kubectl for each matching pod
-            CONTAINER_STATUS=$(kubectl get pod $POD -n $NAMESPACE -o=jsonpath="{.status.containerStatuses[?(@.name=='$CONTAINER_NAME')].state}")
+            CONTAINER_STATUS=$(kubectl get pod $POD -n $NAMESPACE -o=jsonpath="{.status.phase}")
 
             # Check if the container is Running
             if [[ "$CONTAINER_STATUS" == *'running'* ]]; then


michaelefisher:
I have no issue with the change to the status/phase - though I think they might be slightly different in practice.

As for where the normalization files went...we never tested this with BQ v2, and I suspect that your using that destination is impacting things. Do we know how dbt functions with v2? This is going to be a little tough to suss out perhaps without looking through Airbyte code...

Or did you try this with v1 and I'm on the wrong track?

emily-flambe:
> I have no issue with the change to the status/phase - though I think they might be slightly different in practice.
> 
> As for where the normalization files went...we never tested this with BQ v2, and I suspect that your using that destination is impacting things. Do we know how dbt functions with v2? This is going to be a little tough to suss out perhaps without looking through Airbyte code...
> 
> Or did you try this with v1 and I'm on the wrong track?

Fair question - I've tried all of <contents of the PR description> with both the V1 and V2 destinations. (You can see both connections still exist in dev airbyte)

emily-flambe:
ope @michaelefisher I figured out something useful, which is that the BQ destination using GSM is actually on v2 lol

image

I strongly suspect this is why we can't find the Airbyte normalization, and will require some science (not currently urgent) to figure out how that works. Meanwhile @KelseyNewman147 how hard would it be to set up a GSM-pilled BQ destination using v1?

michaelefisher:
Ah ok so what I suspected! 

So, we can set things up locally in Docker and inspect what is going on, but you could also for the sake of extracting dbt, use a different dest connector that supports normalization, I think. The script I wrote looks for any normalization container, so if a connector is running one it'll find it.

emily-flambe:
@michaelefisher word, thank you!

As for this PR, you said it seems fine (even if subtly different) so do you approve of merging it?

PR #259, opened 2023-12-01
fix (adp_payroll): use `safe_cast` when casting string fields to float64

emily-flambe:
on rare occasion (apparently), raw data that comes in as a string that gets regexp_replace'd into a float64 value comes in an unexpected format (maybe blank, or null, or whitespace?) that causes the cast to fail.

I ran this dbt locally and it succeeded 🎉


royconst (approved):
Looks good.

PR #262, opened 2023-12-04
[DE-516] dbt for Facebook Pages

emily-flambe:
spoiler alert this was a very manual process to put together - rip.

Quick note: it's somewhat inconvenient to fetch generated dbt from Airbyte, so I've included all of the generated models in this PR, in their own folder `airbyte_generated`. They are ignored when we run dbt because dbt looks for models in the `models/` folder. 

The approach I took here was basically to use the normalized tables as the data sources. All the dbt is doing is to generate unique hashids for each table, and then the `_base` model delivers all the columns in the table plus `_airbyte_extracted_at` and the `_hashid`. So it's pretty straightforward in a way, but was also a lot of manual fiddling, and only really feasible here because there were only four tables...

Another thing to note (which we already knew, but it bears repeating) is that the dbt normalization we get out of Airbyte when syncing to BQ V1 includes models that break out nested fields into separate tables, which is particularly egregious for fb syncs (both marketing and pages). Here I've opted for what V2 does by default, which is to NOT break out nested fields into their own tables. Thus the dbt for FB Pages only creates four tables: pages, page_insights, posts, and post_insights.

@jitoquinto as the person who has spent the most time thinking about BQ V2 so far, I'd love to chat with you and think about what makes sense for these Phase 2s going forward. 🧐 

### YOU MIGHT BE WONDERING: why not use the actual raw tables from airbyte_internal?

The thing that's medium-annoying about V2 is that because the raw data are all materialized in `airbyte_internal` (which cannot be changed), it is possible but annoying to use those raw tables as the `sources` in dbt projects, because the table names are prepended with the dataset ids. This is "possible" by setting vars within the sql models, but so annoying that I'm declaring it to be Not Worth Pursuing. See this PR for details: https://github.com/community-tech-alliance/dbt-cta/pull/264


dbt-cta/facebook_pages/models/1_cta_full_refresh/page_base.sql
```diff
@@ -0,0 +1,229 @@
...
+    cluster_by = "_airbyte_extracted_at",
+    partition_by = {"field": "_airbyte_extracted_at", "data_type": "timestamp", "granularity": "day"},
+    partitions = partitions_to_replace,
+    unique_key = "_airbyte_page_hashid"
+) }}
+
+-- Final base SQL model
+-- depends_on: {{ ref('page_ab1') }}
+select
+    `_airbyte_page_hashid`,
```
jitoquinto, line 15: is there a reason the column names are wrapped in these ticks `?

jitoquinto (commented):
I left a question about the the marks wrapping the column names but this looks good to me! (note that I didnt look at the airbyte_generated files tho, thats a chomnk!)

And yeah like we talked about earlier, it would be good to set some time to update our script to start doing this if the whole grabbing dbt from airbyte is no longer going to be a viable approach for this in V2. And so hopefully this is less of a manual process!

dbt-cta/facebook_pages/models/1_cta_full_refresh/page_base.sql
emily-flambe: oh yeah, at one point I was hitting an error because the name of a field was a reserved name in BQ (or whatever it's called) - like "group" or something? - and so the SQL wasn't being parsed correctly without backticks to indicate field names. So it seemed easiest to just add the backticks everywhere 🫠

emily-flambe:
> I left a question about the the marks wrapping the column names but this looks good to me! (note that I didnt look at the airbyte_generated files tho, thats a chomnk!)
> 
> And yeah like we talked about earlier, it would be good to set some time to update our script to start doing this if the whole grabbing dbt from airbyte is no longer going to be a viable approach for this in V2. And so hopefully this is less of a manual process!

Wrote a new ticket! https://techallies.atlassian.net/browse/PAD-2150. We can probably reuse a lot of the stuff I wrote back in May, which is actually already in our main branch, we just never use it because it was tailored to a pretty specific use case at the time.

PR #264, opened 2023-12-05
[NOT MEANT FOR SERIOUS REVIEW] use raw tables from `airbyte_internal` for dbt

emily-flambe:
(I deleted the old description for this PR because it was mainly fake news)

OK fun news this actually does work! I did update the `macros` using the ones included with the dbt Airbyte generated, so maybe that did the trick? Anyway, we can in fact use this strat to migrate our existing dbt projects to use V2 destinations, which is good news. @jitoquinto @michaelefisher 

The one thing we will need to modify to get this to run is to pass CTA_DATASET_ID as a dbt var, which I don't think will be challenging at all, but just feels a little silly and will require a code change in the DMDF DAG generation script. This is because profiles.yml can only read env_vars and the SQL model templates can only read vars. :|

To run this locally (per the README in this PR):

## Usage

```shell

dbt run --target cta --select tag:cta --vars '{"cta_dataset_id": "raza_facebook_pages"}'
```

@jitoquinto do you think we should commit this code to the repo so we have it as a future reference for when we finally do migrate existing syncs to V2? Or just document these strats some other way? We are already delivering FB Pages and I'm pretty strongly inclined to continue delivering the pared-down V2 dbt going forward (using the [dbt generation script](https://github.com/community-tech-alliance/dbt-cta/pull/267)).


jitoquinto:
Yo thats awesome! Thanks for figuring this out! Do the changes to the macros break all of our other dbt? I'd say we should probably test that out before merging this in and maybe even leaving it as a Draft PR, like the actblue one, could be good as a reference for future migrations?

> The one thing we will need to modify to get this to run is to pass CTA_DATASET_ID as a dbt var, which I don't think will be challenging at all, but just feels a little silly and will require a code change in the DMDF DAG generation script. This is because profiles.yml can only read env_vars and the SQL model templates can only read vars. :|

I dont think we need to make any changes, I was able to use the `env_var` in the [sql script like this](https://github.com/community-tech-alliance/dbt-cta/pull/251/files#diff-ca07fe329642900e8fe5fe1ff59935e93b8dc59d0544f994c7ac35916d26761dR1) and that seemed to work. So if that's still the case we shouldnt have to make any changes to DMDF since were already passing in the dataset id through env vars?

jitoquinto:
> We are already delivering FB Pages and I'm pretty strongly inclined to continue delivering the pared-down V2 dbt going forward (using the https://github.com/community-tech-alliance/dbt-cta/pull/267).

But also yes, I agree with this and think we should stick with the dbt generation script for V2 too!

emily-flambe:
> Yo thats awesome! Thanks for figuring this out! Do the changes to the macros break all of our other dbt? I'd say we should probably test that out before merging this in and maybe even leaving it as a Draft PR, like the actblue one, could be good as a reference for future migrations?
> 
> > The one thing we will need to modify to get this to run is to pass CTA_DATASET_ID as a dbt var, which I don't think will be challenging at all, but just feels a little silly and will require a code change in the DMDF DAG generation script. This is because profiles.yml can only read env_vars and the SQL model templates can only read vars. :|
> 
> I dont think we need to make any changes, I was able to use the `env_var` in the [sql script like this](https://github.com/community-tech-alliance/dbt-cta/pull/251/files#diff-ca07fe329642900e8fe5fe1ff59935e93b8dc59d0544f994c7ac35916d26761dR1) and that seemed to work. So if that's still the case we shouldnt have to make any changes to DMDF since were already passing in the dataset id through env vars?

Oh!! I guess I didn't try hard enough to pass the env_var into my sql models, so I'm glad you figured it out!

> Do the changes to the macros break all of our other dbt? 

I will do some poking and find out!

emily-flambe:
@jitoquinto I took the updated macros in this PR for a spin on Empower and FB Marketing in dev (locally) and they worked fine! Ill close this PR, open a separate PR to update the dbt macros, and then we can use your earlier PR as a guide for updating the dbt for V2. https://github.com/community-tech-alliance/dbt-cta/pull/251/files#diff-ca07fe329642900e8fe5fe1ff59935e93b8dc59d0544f994c7ac35916d26761dR1

PR #265, opened 2023-12-05
feat: deliver elementary data to `{CTA_DATASET_ID}_elementary`

emily-flambe:
I thought this wasn't working because running locally didn't appear to write any tables to elementary at the intended new destination, but it DOES seem to work when run on composer, so I guess this strat really does work after all?

NOPE it doesn't

really just ignore this PR ok ty


emily-flambe:
I can't get this to work lol rip

PR #269, opened 2023-12-11
[DE-492] remove dbt for extra GA4 models (not needed)

emily-flambe:
Per discussion in the GA4 Phase 1 ticket, CTA will only be offering the GA4 data streams that Airbyte enables by default, which is to say all of the streams that do not end in `_report`. This PR removes the dbt for those models. cc @royconst 

This PR might be easier to grok by looking at the files that are left after all the deletions:
image

`_report` tables have been removed from sources.yml as well.


jitoquinto (approved):
oh thats a lot of deletions 😮

kanelouise (commented):
bye bye files

PR #272, opened 2023-12-13
[DE-492] google_analytics_ga4: add tests back in and run sqlfluff

emily-flambe:
forgot to add tests back in after yeeting all the models for the `_reporting` tables, oops. Also ran SQL fluff to beautify ✨


jitoquinto (approved):
:catsalute:

PR #270, opened 2023-12-12
feat: update macros from Airbyte generated dbt (Dec 8 2023)

emily-flambe:

PR #273, opened 2023-12-19
Revert addition of should_full_refresh macro

emily-flambe:
Reverting this file added in https://github.com/community-tech-alliance/dbt-cta/pull/270/files. It forces dbt to run full refresh if a table does not contain the field `_airbyte_ab_id`. Rude!


emily-flambe:
YOLO-merging to fix a bug affecting syncs in prod.

PR #278, opened 2023-12-28
[DE-554] add dbt models for Sprout Social

emily-flambe:
This sync uses BQ V2 destination in Airbyte, so I initialized the dbt using `utils/generate_dbt/generate_dbt.py`. I then did a little extra work to unnest all the metrics and dimensions fields for the post analytics and profile analytics models. (Note @kanelouise that we will need to do something similar when we add Twitter to the mix, but it shouldn't be that much additional work.)

🎉


kanelouise (approved):
Love the unnesting! This will make it much nicer for report building. Thanks, @emily-flambe!

PR #283, opened 2024-01-07
feat: automatically generate deduping CTEs when generating dbt from BQ V2 tables

emily-flambe:
This PR adds a step to the code used to generate dbt files from tables in BQ (created from using the BQ V2 destination in Airbyte) to generate additional CTE models for each model that dedupe on the `_hashid` field generated in the `_ab1` models. This deduplicates data in cases where duplicate rows exist in the source data (which is rare, but irritating - and now we don't need to worry about it anymore!)


utils/generate_dbt/generate_dbt.py
```diff
@@ -96,7 +104,8 @@ def generate_sql_files(
         columns_for_hashid = [
             f"'{field[0]}'"
             for field in columns
-            if field[1] in ["STRING", "INT64", "FLOAT64", "BOOL", "DATE", "TIMESTAMP"] and not field[0].startswith("_airbyte")
+            if field[1] in ["STRING", "INT64", "FLOAT64", "BOOL", "DATE", "TIMESTAMP"]
```
jitoquinto, line 107: We should probably think about how to make this work for the nested and repeated field types. Definitely not a rn thing, but i think a possible edge case is that we could have a table where all of the fields we are creating the hash id with could be the same for multiple records but then they have unique columns that were just not including in the hash field (json cols). So it would count it as a dupe when not really a dupe. Does that make sense? or am i overthinking this?

utils/generate_dbt/templates/0_ctes_ab2.sql
```diff
@@ -0,0 +1,11 @@
+-- ensures the base model contains only one row per hashid
+-- this deduplicates data even if the source data contains duplicate rows
+
+SELECT * FROM 
+(
+SELECT 
+    *,
+    ROW_NUMBER() OVER (PARTITION BY _airbyte_{{ table }}_hashid ORDER BY _airbyte_extracted_at desc) as rownum 
+from {% raw %}{{ ref('{% endraw %}{{ table }}{% raw %}_ab1') }}{% endraw %}
```
jitoquinto, line 9: wowow templates! so cool!

jitoquinto (approved):
Left a comment but lgtm!

utils/generate_dbt/templates/0_ctes_ab2.sql
emily-flambe: ikr

PR #284, opened 2024-01-07
create blocks_incremental models with deduping CTEs

emily-flambe:
Lots of files changed here, which makes it hard to grok what's going on. Here's what I did that led to this PR:

1) deleted all the existing dbt files, which were originally generated by Airbyte when using the BQ V1 destination connector.
2) generate new dbt files using utils/generate_dbt/generate_dbt.py.

TODO before this is ready for review:
3) The dbt generation script creates both full_refresh and incremental models for each table, so I need to delete the models as needed so that each one only appears once according to their respective sync strategy
4) Add tests
5) SQL Fluff


emily-flambe:
closing this PR to open separate ones that will make it easier to see what the new models look like

PR #286, opened 2024-01-07
add new blocks_incremental models

emily-flambe:
This PR will ultimately replace the current dbt models for `blocks_incremental` with new ones created by running generate_dbt.py on tables created using Airbytes V2 destination for BigQuery. The PR compares against a branch that has removed all of the existing models, which makes it easier to see the new models added here without the visual clutter of seeing all the file changes and deletions when diffing against the original set of dbt models generated by Airbyte using the BQ V1 connector.

Steps that led to the code on this branch:

1) deleted all the existing dbt files, which were originally generated by Airbyte when using the BQ V1 destination connector. (This change is already reflected in ec/delete-blocks-incremental, the branch this PR is diffing against.)
2) generated new dbt files using utils/generate_dbt/generate_dbt.py.
3) The dbt generation script creates both full_refresh and incremental models for each table, so I deleted the models as needed ([this commit](https://github.com/community-tech-alliance/dbt-cta/pull/286/commits/ea5f1d20136c259418c6f12e7add743257dc551c)) so that each one only appears once according to their respective sync strategy

~TODO before this is ready for review:~ These are all completed and this PR is ready for review 🎉 .
- [x] Add tests
- [x] SQL Fluff
- [x] run `dbt run` and `dbt test` locally

TODO after merging:
- [ ] merge ec/delete-blocks-incremental into main (https://github.com/community-tech-alliance/dbt-cta/pull/285)


dbt-cta/blocks_incremental/models/1_cta_incremental/_1_cta_incremental__models.yml
```diff
@@ -0,0 +1,4556 @@
...
+      - name: smarty_streets_match_data
+        description: ''
+      - name: scan_id
+        description: ''
+      - name: attempted
+        description: ''
+      - name: _airbyte_registration_forms_hashid
+        description: ''
+        tests:
+          - unique
```
jitoquinto, line 139: Do we want to have 2 uniqueness tests per table? wondering if just keeping 1 would be good enough especially considering how many tables there are

jitoquinto (approved):
I did not look through all 600 files 😓 but looked through a handful of models in the different folders + tests and sources yamls which all look good to me! I left a question regarding tests for incremental model but not a blocker.

dbt-cta/blocks_incremental/models/1_cta_incremental/_1_cta_incremental__models.yml
emily-flambe: yeah, I did make the deliberate decision to add tests for `id` in addition to `hashid` even though in theory they are duplicative - my inclination is to leave them in for now and remove later if Composer is having trouble executing the dbt in a reasonable amount of time 🫠

PR #290, opened 2024-01-23
WIP: fixes to Sprout Social sync

emily-flambe:


dbt-cta/sprout_social/models/0_ctes/facebook_post_analytics_ab3.sql
```diff
@@ -0,0 +1,11 @@
...
+-- this deduplicates data even if the source data contains duplicate rows
+
+SELECT * EXCEPT (rownum) FROM 
+(
+SELECT 
+    *,
+    ROW_NUMBER() OVER (PARTITION BY _airbyte_facebook_post_analytics_hashid ORDER BY _airbyte_extracted_at desc) as rownum 
+from {{ ref('facebook_post_analytics_ab2') }}
+)
+where rownum=1
```
kanelouise, line 11: I'm curious why this line`where rownum =1` is the exception. Also just noticing that this is happening in `ab_3` not `ab_2` like the other `where rownum=1` commands.

kanelouise (approved):
I just have one q, but it's more for my edification than anything. Don't see anything sus.

dbt-cta/sprout_social/models/0_ctes/facebook_post_analytics_ab3.sql
emily-flambe: > I'm curious why this linewhere rownum =1 is the exception

Not quite sure I understand the question here - can you help me understand what you mean that it is the exception?

The reason these are in _ab3 instead of _ab2 is that this particular sync needs an additional CTE in the middle to calculate the `_hashid` fields, since some of the values used to calculate those come from unnesting the `metrics` fields - so the first CTE needs to run and do that unnesting before the second CTE uses those metrics to calculate the hashid. Then the `_ab3` CTE does the deduping by _hashid. Does that make sense?

(You might be wondering why we are bothering to do the unnesting for this data since we've talked about generally not doing that - in this instance it was easy for me to do and there aren't that many tables, so I went ahead and did it to make the analysts lives easier when setting up the dashboards using this data 👀 )

kanelouise: Lol makes sense that you didn't understand what I was saying because I was just confused. Anyway, after talking through this in our 1:1 earlier I have more clarity!

PR #293, opened 2024-01-24
feat: new script for generating _ab4 deduping CTEs

emily-flambe:
This PR adds a quick and dirty script that iterates through the tables in a project (determined via user input) and generates `0_ctes/TABLE_ab4.sql` files that dedupe rows using the `_hashid` field in each table.

Because this is quick and dirty, the script hardcodes the template being used to generate the SQL, and assumes that the desired filename has the suffix `_ab4`. This should be fine for most or all of our syncs that still do not have these CTEs included, and the main purpose of this script is to facilitate adding those models to existing projects. Future syncs will use `generate_dbt.py` to generate dbt models, and that script already creates deduping CTEs, so this script will no longer be useful once all existing projects have had these models added.

### DIFF
--- a/utils/generate_dbt/generate_deduping_ctes.py
+++ b/utils/generate_dbt/generate_deduping_ctes.py
@@ -21,7 +21,7 @@ def main():
     folder_path = f"../../dbt-cta/{sync_name}/models/0_ctes"
 
     # Read the template from the separate .sql file
-    with open("templates/0_ctes_ab2.sql", "r") as template_file:
+    with open("templates/0_ctes_ab4.sql", "r") as template_file:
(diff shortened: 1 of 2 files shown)


kanelouise (approved):
LGTM!

kanelouise (commented):
Ooo, I guess we could add a loop for updating the base tables as well?

emily-flambe:
> Ooo, I guess we could add a loop for updating the base tables as well?

oh that's a good idea! bet that would be easy, I'll ask chatgpt to throw that in there lol

kanelouise (commented):
Cool, love the base tables addition.

PR #294, opened 2024-01-24
feat (action network): add _ab4 CTEs to dedupe on _hashid fields

emily-flambe:
We should add these CTEs to all of our dbt projects, but this change is a priority for Action Network data because the Airbyte source connector (Redshift) sometimes only does a partial sync and needs to restart, resulting in duplicate rows in the raw tables that need to be handled downstream (via these models).


jitoquinto:
That makes sense! Do the base models have to be updated to pull from the new cte?

emily-flambe:
> That makes sense! Do the base models have to be updated to pull from the new cte?

yeah lol whoops! https://github.com/community-tech-alliance/dbt-cta/pull/294/commits/dcffdca726363c3dcbd016b1f92102c0b133b641

kanelouise (approved):
With the base update, looks good!

PR #303, opened 2024-01-29
POC of dbt for delivering data via Analytics Hub

emily-flambe:
This PR does the following in order to enable delivering data via Analytics Hub subscriptions:

1 - creates a new profile in profiles.yml:

```yaml
    cta_delivery:
      project: "{{ env_var('CTA_PROJECT_ID') }}"
      dataset: "{{ env_var('CTA_DATASET_ID') }}_delivery"
```

2 - adds a new model type to dbt_project.yml, which materializes data as tables using full_refresh:

```yaml
    2_cta_delivery: # for delivery via Analytics Hub subscriptions
      +tags: 
        - cta_delivery
      +materialized: table
      +full_refresh: true
```

3 - Adds _bunch_ of new models that will enable us to deliver data via Analytics Hub subscriptions. These are all `SELECT * FROM {base_table}` of the exact same form that we currently use to deliver materialized views.

I don't think we need to save this anywhere, but I used [this shell script](https://gist.github.com/emily-flambe/260c616623ac3f458ad8dbf10187342b) to iterate through all the folders and create new models for `2_cta_delivery`. The models all have the prefix `cta_` because dbt throws an error if multiple models exist in a project using the same name, even if they are contained in different subfolders.

This is currently deployed to dev and is being used in DAGs that are POC-ing the new Analytics Hub data delivery method (including, for now, sprout_social and actblue - more to come soon).


jitoquinto:
This makes sense to me! One question though, there are a few vendors that we have `3_partner_views` how should we handle that in this new world of read only datasets? Should those be made into their own tables too?

emily-flambe:
> This makes sense to me! One question though, there are a few vendors that we have `3_partner_views` how should we handle that in this new world of read only datasets? Should those be made into their own tables too?

Hmm, yeah those are really a relic of a bygone era. Probably easiest to just move their SQL into `_delivery` tables as well. There's no particular reason why they were ever `views` instead of matviews, I don't think. I'll do that now 👀

emily-flambe:
> cb998b3



> > This makes sense to me! One question though, there are a few vendors that we have `3_partner_views` how should we handle that in this new world of read only datasets? Should those be made into their own tables too?
> 
> Hmm, yeah those are really a relic of a bygone era. Probably easiest to just move their SQL into `_delivery` tables as well. There's no particular reason why they were ever `views` instead of matviews, I don't think. I'll do that now 👀

yay! https://github.com/community-tech-alliance/dbt-cta/pull/303/commits/cb998b3099f1d53c32a8d6af5d5a43b15deeb64b

jitoquinto:
> Hmm, yeah those are really a relic of a bygone era. Probably easiest to just move their SQL into _delivery tables as well. There's no particular reason why they were ever views instead of matviews, I don't think. I'll do that now 👀 

sounds good! One other thing i found was that some of the matview models (and thus the AH delivery models) have a config that set `full_refresh = false`. So for these, the full_refresh at the project yaml will not go through cause i think model configs have a higher precedence

emily-flambe:
> > Hmm, yeah those are really a relic of a bygone era. Probably easiest to just move their SQL into _delivery tables as well. There's no particular reason why they were ever views instead of matviews, I don't think. I'll do that now 👀
> 
> sounds good! One other thing i found was that some of the matview models (and thus the AH delivery models) have a config that set `full_refresh = false`. So for these, the full_refresh at the project yaml will not go through cause i think model configs have a higher precedence

oh dang, that's a good catch. Seems to me like there's no real reason to keep those configs included in any of the `delivery` or `matviews` models, so I'll just yeet them entirely - unless you know of any good reason they should exist in some cases?

emily-flambe:
> > > Hmm, yeah those are really a relic of a bygone era. Probably easiest to just move their SQL into _delivery tables as well. There's no particular reason why they were ever views instead of matviews, I don't think. I'll do that now 👀
> > 
> > 
> > sounds good! One other thing i found was that some of the matview models (and thus the AH delivery models) have a config that set `full_refresh = false`. So for these, the full_refresh at the project yaml will not go through cause i think model configs have a higher precedence
> 
> oh dang, that's a good catch. Seems to me like there's no real reason to keep those configs included in any of the `delivery` or `matviews` models, so I'll just yeet them entirely - unless you know of any good reason they should exist in some cases?

@jitoquinto I just cleared out all of the model-level configs I found - these were present in `square` and `spoke` (which are not currently in use) as well as `stackadapt` and `usvote_foundation`. I can't think of any reason to keep them for any of those syncs 🤔

jitoquinto:
> oh dang, that's a good catch. Seems to me like there's no real reason to keep those configs included in any of the delivery or matviews models, so I'll just yeet them entirely - unless you know of any good reason they should exist in some cases?

nah i think its fine to yeet them! oh one other thing i saw (sorry i had to pull down locally cause gh was just not loading the diff 🙃 ) on the project.yml could you remove the `+full_refresh: true` in the `2_partner_matview` config?

emily-flambe:
> > oh dang, that's a good catch. Seems to me like there's no real reason to keep those configs included in any of the delivery or matviews models, so I'll just yeet them entirely - unless you know of any good reason they should exist in some cases?
> 
> nah i think its fine to yeet them! oh one other thing i saw (sorry i had to pull down locally cause gh was just not loading the diff 🙃 ) on the project.yml could you remove the `+full_refresh: true` in the `2_partner_matview` config?

hmm, idk if I'm being dense or Github is being weird, but I don't see that config present on my branch:

image

jitoquinto:
oh, yeah my local was out of date! can ignore that request!

kanelouise (approved):
Exciting and computer melting changes are afoot! I followed along on y'all's journeys of discoveries and I don't have anything else to add.

jitoquinto:
yeah my computer keeps freezing when i try to go to the changes page to approve 😭 but i will leave my approval here as a comment

PR #305, opened 2024-01-30
[DE-640] feature: add deduping cte's to actblue

Frydafly:


jitoquinto:
hmm, i wanna say these models already have deduping in the `0_ctes/*_ab3` files

dbt-cta/actblue/models/sources.yml
```diff
@@ -66,9 +66,6 @@ sources:
         tests:
           - dbt_expectations.expect_column_values_to_be_of_type:
               column_type: timestamp
-      - name: _airbyte_data
-        tests:
-          - not_null
   - name: paid_contributions_stream
     columns:
       - name: _airbyte_ab_id
```
emily-flambe, line 71: ope, @Frydafly this is not your fault because you inherited this 🙃  but can we change all the tests for `_airbyte_ab_id` and make them instead test uniqueness on the `_hashid` fields for each table?

emily-flambe (changes requested):
Requested a small tweak to the schema yaml containing the dbt tests!

emily-flambe:
> hmm, i wanna say these models already have deduping in the `0_ctes/*_ab3` files.
> 
> https://github.com/community-tech-alliance/dbt-cta/blob/main/dbt-cta/actblue/models/0_ctes/cancelled_recurring_contributions_stream_ab3.sql#L37
> 
> Update - Oh i guess only the `cancelled_recurring_contributions_stream` has that 🤔

oh good catch, I'd forgotten about that. @Frydafly can you also revise `cancelled_recurring_contributions_stream_ab3` so that it does NOT include the window function to do the deduping, and instead we do that in _ab4 to keep things consistent with the other Actblue models?

emily-flambe (approved):
LGTM now! ty!

jitoquinto (approved):
LGTM too! :shipit:

PR #314, opened 2024-01-30
feat: sprout social base models should be incremental

emily-flambe:
The Airbyte source always overwrites the past 365 days (because that's the max lookback window enabled by the API) - thus the dbt should run incrementally to merge in new rows, preserving rows that were previously synced.

(This PR also includes proposed changes to the dbt project to create `_delivery` tables in the CTA project for Analytics Hub delivery - see https://github.com/community-tech-alliance/dbt-cta/pull/303)


dbt-cta/dbt_project.yml
```diff
@@ -39,7 +39,12 @@ models:
         - cta
       +materialized: incremental
       +on_schema_change: sync_all_columns
-    2_partner_matviews:
+    2_cta_delivery: # for delivery via Analytics Hub subscriptions
+      +tags: 
+        - cta_delivery
+      +materialized: table
+      +full_refresh: true
```
jitoquinto, line 46: oh nice! does this mean we can actually use full_refresh cause we're not making matviews off these tables?

dbt-cta/sprout_social/models/0_ctes/facebook_post_analytics_ab3.sql
```diff
@@ -1,7 +1,7 @@
 -- ensures the base model contains only one row per hashid
 -- this deduplicates data even if the source data contains duplicate rows
 
-select * except (rownum) from
+select * except (rownum) except (rownum) from
```
jitoquinto, line 4: this one has an extra `except`

dbt-cta/sprout_social/models/0_ctes/facebook_profile_analytics_ab3.sql
```diff
@@ -1,7 +1,7 @@
 -- ensures the base model contains only one row per hashid
 -- this deduplicates data even if the source data contains duplicate rows
 
-select * except (rownum) from
+select * except (rownum) except (rownum) from
```
jitoquinto, line 4: ++ extra except

dbt-cta/sprout_social/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
jitoquinto, line 1: Should these tests be moved over to the 1_cta_incremental folder?

jitoquinto (changes requested):
:catjam: exciting to see analytics hub transition starting! It looks like some extra `except`s made their way in there but other than that this lgtm

dbt-cta/sprout_social/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
emily-flambe: ope! yes! ty for catching

dbt-cta/sprout_social/models/0_ctes/facebook_profile_analytics_ab3.sql
emily-flambe: ope!! fixed! https://github.com/community-tech-alliance/dbt-cta/pull/314/commits/a6be26c9fb3ac95fe2e14fe2393d02283b0b602a

dbt-cta/dbt_project.yml
emily-flambe: yeah! full refresh all the way baby! pretty sure that doesn't break anything in the linked datasets - probably just a very brief blip of unavailability while each table is rebuilding every morning.

dbt-cta/dbt_project.yml
```diff
@@ -39,7 +39,12 @@ models:
         - cta
       +materialized: incremental
```
kanelouise, line 40: The `1_cta_incremental` being incremental to the `2_cta_delivery` is breaking my brain a little. I'd love to check my understanding for it here:

In the `1_cta`'s (both incremental and full refresh) the data is pulled from the raw AB tables. Then, per the new AH magic, the `1_cta`'s (both incremental and full refresh) are fully recreated because the `1_cta_full_refresh` data is all of the data ever, every single time it's run; and the `1_cta_incremental` is an added to each time it's run... therefore the full_refresh works regardless of how the data is being pulled from AB. i don't know if those words make sense to anyone but me... lol.

emily-flambe: @kanelouise, so the new Analytics Hub related changes don't make any changes to how the `1_cta_` models work. The way we've had those set up is that `1_cta_full_refresh` does a truncate-and-rebuild of base tables, while `1_cta_incremental` only adds in new rows based on the `unique_key`.

The Sprout Social API only allows for us to sync the past 365 days of data on each run, right? So if we update the base tables using `1_cta_full_refresh`, the final data will only ever include the past 365 days of data. By changing this to `1_cta_incremental` instead, then the dbt won't get rid of the existing data, but will instead continue to merge in new rows.

This PR got confusing because it should have ONLY included the change to make the Sprout Social base models be incremental instead of full refresh, but in order to keep dev deploys from yeeting my Analytics Hub POC work, I included the change to add 2_cta_delivery tables as well. But does all of this make sense now?

PR #316, opened 2024-02-01
feat: add BQ labels

emily-flambe:
Adding these labels will make it easier for us to determine how many and which data objects in partner projects were created by CTA versus created by the partner.


emily-flambe:
this branch and PR got messy with conflicts and it only really needs to change one file, so I'm gonna close this PR and open a new one

PR #318, opened 2024-02-02
feat: add BQ labels to dbt models (attempt #2)

emily-flambe:
Adding these labels will make it easier for us to determine how many and which data objects in partner projects were created by CTA versus created by the partner.

The[ first PR attempting to add this in](https://github.com/community-tech-alliance/dbt-cta/pull/316) became riddled with conflicts, and this is a tiny change so it was faster to just abandon that branch and open this new PR instead.


emily-flambe:
@michaelefisher I tried using this on mailchimp data in dev (running locally with --full-refresh) and no labels appear in BigQuery :( so I'm not actually sure this strat is going to work...

image

https://console.cloud.google.com/bigquery?project=dev3869c056&ws=!1m5!1m4!4m3!1sdev3869c056!2spartner_a_mailchimp!3scampaigns

jitoquinto (approved):
:shipit:

emily-flambe:
I've tried running this locally a few times now, both with `--full-refresh` and after fully dropping tables in BQ, and this change doesn't appear to be successfully applying labels. And this isn't actually that important, so I'm abandoning this PR for now. RIP

PR #319, opened 2024-02-02
feat: dbt generation script creates 2_cta_delivery models

emily-flambe:
Pursuant to the changes in [this PR](https://github.com/community-tech-alliance/dbt-cta/pull/303) adding the model type `2_cta_delivery` for future delivery using Analytics Hub linked datasets, this PR updates our script for generating dbt for new syncs as follows:

1 - a step is added that creates SQL models for `2_cta_delivery`. These are named based on the name of the original table. (Previously that was the case for materialized views.)
2 - materialized views now have the suffix `_mv`. Going forward, these will not be delivered to partners for new syncs by default, but we want to keep these models around until/unless we completely deprecate materialized views.
3 - also made a tweak to the template for `1_cta_full_refresh` to stay consistent with our new standard practice of replacing date partitions for the past 6 days - which is probably unnecessary now that we run a pre-hook to TRUNCATE each table before running the dbt models, but we might as well be consistently unnecessary.


jitoquinto (approved):
I guess we still need the extra date partition replacements 🙃

Frydafly (approved):
Screenshot 2024-02-05 at 1 15 13 PM
ran for sosha!

PR #320, opened 2024-02-02
feat: update dbt models for Action Builder

emily-flambe:
NOTE: this needs to be reviewed and merged in tandem with https://github.com/community-tech-alliance/airflow-dags/pull/636 or the sync will fail in prod.

---

While working on enabling Analytics Hub delivery for Action Builder, I found that our dbt is hideously out of of date, and still assumes the raw data is being synced by Airbyte (which it is not).

In order to run dbt tests on the `_base` tables, which are materialized by the DAG rather than by dbt, we need to run source tests by adding test conditions to `sources.yml` - which is challenging to view among all the file diffs, so you can check it out [here](https://github.com/community-tech-alliance/dbt-cta/blob/77ccd4578eba037b7268a14de80ae5e0029fd274/dbt-cta/action_builder/models/sources.yml).

This PR also creates `2_cta_delivery` and, for good measure, `2_partner_matviews`.

`2_cta_delivery`: these are necessary for delivering Action Builder tables via Analytics Hub.
`2_partner_matviews`: these are included for consistency with other models in this repo; currently the DAG materializes matviews without using dbt, but these models would make it possible for us to (optionally) provide matviews in addition to AH linked datasets in the future.

This dbt is succeeding in dev composer and also succeeds when I run locally using one of our datasets in prod.


dbt-cta/action_builder/models/sources.yml
```diff
@@ -3,137 +3,365 @@ sources:
...
+          - unique
+          - not_null
   - name: emails_base
+    columns:
+      - name: id
+        tests:
+          - unique
+          - not_null
   - name: entities_base
-  - name: entity_connection_types_base
```
jitoquinto, line 106: It looks like this table is getting removed completely, is that expected?

dbt-cta/action_builder/models/sources.yml
```diff
@@ -3,137 +3,365 @@ sources:
...
+          - unique
+          - not_null
   - name: tag_categories_base
+    columns:
+      - name: id
+        tests:
+          - unique
+          - not_null
   - name: tag_groups_base
-  - name: taggable_logbook_base
```
jitoquinto, line 131: ++ to this table

jitoquinto (approved):
Oof this was really out of date! Thanks for cleaning up and getting it ready for the brave new world! Only thing I saw is that it looks like a couple of tables are getting removed so just wanted to make sure that was expected. If it is then :shipit:

dbt-cta/action_builder/models/sources.yml
emily-flambe: oh, yeah so an annoying property of the Action Builder sync is that for one partner there are a few tables that we cannot sync because of data replication conflicts with the database. But since the syncs are all sharing the same dbt project, the dbt yaml has to work for all of them. And if we include references to tables that don't exist, the dbt will fail. We could create empty versions of those tables on an as-needed basis, though, and that might be the more reasonable way to handle this... I shall contemplate

emily-flambe: COOL FACT, since the DAG runs all the table models individually, it actually doesn't seem to mind that sources.yml includes sources that might not be present for a particular partner!

jitoquinto: ooh that is a cool fact!

jitoquinto (approved):
thanks for investigating those tables! lgtm 👍

PR #321, opened 2024-02-05
fix: remove `TRUNCATE` pre-hook

emily-flambe:
This was a neat idea but it makes the dbt fail if the base table doesn't already exist, and there's no clean way to make it only run if the table already exists. RIP in spaghetti never forgetti


jitoquinto (approved):
👋

PR #322, opened 2024-02-06
[DE-552] create soSha sync - phase 2

Frydafly:


emily-flambe (approved):
I already ran this locally and it seems good to me! :shipit:

PR #329, opened 2024-02-14
[DE-657] feat: add twitter models to sprout social

Frydafly:
to do: 

- [x] unnest metrics fields
- [x]  adding surrogate keys to ab2 based on all fields in ab1
- [x] dbt run
- [x] add yaml for twitter models


emily-flambe (commented):
So far so good, looks like the schema yaml still needs tests for the two new tables added (which is in your todo list already)! Thanks so much also for updating the `_ab3` references in the other analytics tables (oops 🫠)

emily-flambe (approved):
:shipit:

PR #336, opened 2024-03-06
fix: add data types to non-string fields for mobile commons sync

Frydafly:
related Jira ticket: https://techallies.atlassian.net/browse/DE-717 

Mobile Commons did not pass QA due to timestamps showing up as strings. 

In this PR, I converted these fields to timestamps, and also converted integer and boolean fields to the proper data types.


Frydafly:
> This looks good to me! Has this been ran and tested in dev yet? Also what are your thoughts on [safe_cast](https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_functions#safe_casting)? I know I've used that in some other syncs since null values would cause failures and cause the whole dbt run to fail. But i can also see safe casting causing silent failures and causing there to be null data instead of notifying us of bad data.

@jitoquinto yes, it ran successfully in dev. 
here's one matview 

```sql
SELECT * FROM `dev-partner-a-25693fa5.mobile_commons.broadcasts_mv` LIMIT 1000
``` 

as for safe_cast, i'm happy to do either, depending on your preferences. cast ran successfully with the data, but let me know and can switch it over to avoid headaches in the future.

PR #339, opened 2024-04-02
updates for migration to v2 destination

anneram:


anneram:
> As discussed in Slack, while we're making this change could you please swap out the suffix `_stream` for `_base` so that these models are consistent with our other syncs? 🙏 yay 🙏

@emily-flambe I think I got everything but still learning the ropes around here so would love someone to double check!

emily-flambe (approved):
:shipit: !

PR #342, opened 2024-04-22
hotfix (blocks): registration_type in scans_qc_overview is now a string

KelseyNewman147:
@emily-flambe fixed this field on another table a few weeks ago in [this PR](https://github.com/community-tech-alliance/dbt-cta/pull/338), but the same change is needed on the `scans_qc_overview` table now too.


kanelouise (approved):
👍

PR #343, opened 2024-05-02
Migrate Mobilize to BQ v2

anneram:
While I was making the updates listed in the [migration doc](https://docs.google.com/document/d/1KgTb2TonitMItHch_4CfbBCtPNWELLdZ-e746S_-GQA/edit) I noticed all the `0_ctes/ab2` models had a hard-coded ab1 CTE instead of referencing the actual (identical) `ab1` model, and the `ab3` models had both ab1 and ab2 CTEs hard-coded instead of referencing those models. I assume this was just an older/previous dbt strategy and these were overlooked/forgotten about, so I went ahead and just cleaned up these models to use the proper references. Everything runs successfully!

###

`airbyte-configs` PR: https://github.com/community-tech-alliance/airbyte-configs/pull/138
`airflow-dags` PR: https://github.com/community-tech-alliance/airflow-dags/pull/731


emily-flambe (approved):
Looks like this ran successfully in dev! I only checked the matview for `affiliations` but it looks good to me :shipit:

PR #346, opened 2024-05-10
[DE-660] create models for thrutext

emily-flambe:
Accompanies this PR adding a new Thrutext DAG: https://github.com/community-tech-alliance/airflow-dags/pull/741

Currently we are only syncing 3 tables in this sync, which makes the dbt fairly straightforward. The DAG truncates and rebuilds the `_stg_*` tables on each run, and the dbt incrementally appends rows to the `*_base` tables, which form the basis of matviews delivered to the partner project.

Different partners could conceivably sync different tables in the future, but the DAG only runs the dbt models specific to the table in each individual task, so it should work fine if we ever add models for tables that some partners don't need. (This is how Openfield works too.)

This dbt runs successfully locally and the dev DAG is succeeding too 🎉


emily-flambe:
Moving this back to draft status while I tinker with using SELECT * and `on_schema_change: add` and delivering this via Analytics Hub instead of matviews (to capture possible custom fields in the future)

emily-flambe:
> Left some comments about the VAN specific columns. Im pretty sure those are custom fields for the partner. I wanna say i used this list to create the schema files for the base columns that all thrutext customers will have
> 
> https://help.thrutext.io/support/solutions/articles/44001444138

Ope, good catch, thanks! I'll fix the hashid columns so that they only use fields common to all customers

emily-flambe:
YOLO-merging to get the sync to succeed in prod - I'm sure it's probably fine

PR #352, opened 2024-05-20
increase `job_execution_timeout_seconds` 600->1200 for `cta_1_thread` bq profile

emily-flambe:
We are starting to experience dbt task failures in Composer for the Blocks sync, which contains over 150 models. Maybe it hits the timeout because it's only using one thread? Who knows? Increasing the timeout should at least get the sync to succeed. (Hopefully - I'm running the dbt locally now and will update)

### DIFF
--- a/dbt-cta/profiles.yml
+++ b/dbt-cta/profiles.yml
@@ -46,7 +46,7 @@ default:
       threads: 1
       type: bigquery
       job_creation_timeout_seconds: 30
-      job_execution_timeout_seconds: 600
+      job_execution_timeout_seconds: 1200
       job_retries: 5
       job_retry_deadline_seconds: 1200
     partner:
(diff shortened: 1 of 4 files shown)


emily-flambe:
> 🫡 Do we know if the query is taking this long? or is it just that we are submitting a bunch of queries and it takes a long time for it even run

My hunch is that the issue isn't any single query that's running for a long time, but rather the fact that we are submitting a whole bunch of BQ jobs at the same time. If you look at the error logs for each individual failed task from this morning, the tables that failed were different tables, until finally the job succeeded. Which I think might mean that on each run, some of the queries did successfully merge in new rows, and eventually it was able to get through all of them? You'd think if there was one query that was terrible, we would see the same error message repeatedly, and also it would be weird for the sync to have eventually succeeded. 🧐

PR #353, opened 2024-05-20
thrutext: on_schema_change should be `sync_all_columns`

emily-flambe:
This enables the dbt to run if columns are removed. Having this set to `add` only allows new columns to be added, but sometimes columns disappear in a Thrutext export.


huystuhh (approved):
:shipit:

PR #356, opened 2024-05-28
switch incremental Action Network tables to `1_cta_incremental` models

emily-flambe:
All but 3 of the Action Network tables now sync incrementally (via Airbyte), which means the date replacement strat we use for `1_cta_full_refresh` models (to avoid dropping the table) results in duplicate rows appearing after 7 days.

Only 3 of the AN tables should be run with full refresh: `core_fields_counties`, `core_fields_ocdids`, and `field_names_groups`. dbt models for all other tables have been moved to 1_cta_incremental. The SQL is the same, but omits the stuff at the beginning and end of the `1_cta_full_refresh` models to do the date partition replacement.

Also probably worth noting that I've changed the `unique_key` field in the incremental models to be `id`. That is the primary key for all incremental tables, and it's the field we are using to dedupe rows in the `_ab4` models, so it keeps things internally consistent by naming it as the `unique_key` in the `_base` models.

I've been testing this dbt locally using a copy of prod data (copied to `partner_a_action_network`) - still need to test in the dev DAG.


dbt-cta/action_network/models/1_cta_full_refresh/_1_cta_full_refresh__models.yml
```diff
@@ -1,77 +1,18 @@
-version: 2
+3version: 2
```
kanelouise, line 1: This `3` may be a typo.

dbt-cta/action_network/models/1_cta_full_refresh/core_fields_counties_base.sql
```diff
@@ -13,6 +13,7 @@
     partitions = partitions_to_replace,
     unique_key = "_airbyte_raw_id"
```
kanelouise, line 14: Is this still ok as the `unique_key` or do we want it to be `id` like the `incremental` configs?

kanelouise: (same with the other 2 `full_refresh`)

kanelouise (commented):
Looks good, thanks for getting this out so quickly. I have just a few questions!

dbt-cta/action_network/models/1_cta_full_refresh/core_fields_counties_base.sql
emily-flambe: Oh yeah! good catch - these do have primary keys (or at least fields called `id`) but we aren't syncing them incrementally in Airbyte because they don't have cursor fields. So yep I'll make these unique_keys be `id` as well (though since these are full refresh I think the unique_key doesn't actually impact dbt's behavior)

emily-flambe: https://github.com/community-tech-alliance/dbt-cta/pull/356/commits/4f3a057a85b16b4cfabe24d507e30fb8c45cd865

emily-flambe:
@kanelouise this is succeeding in dev Airflow, and I also ran `dbt run` and `dbt test` locally using data in prod and everything is passing. This is ready for your (hopefully) final consideration 🫡

kanelouise:
> oof good to know past us had a real reason to have this different strat for incremental!


true, @jitoquinto, but past y'all also had the wisdom of buying us time with the seven days of timestamp truncation!

kanelouise (approved):
LGTM, @emily-flambe!

PR #363, opened 2024-06-07
hotfix: yeet unnecesary columns to create cta_hashid

kanelouise:
The dbt was failing to create a `_cta_hashid` because there were optional fields included. This PR pares down the amount of columns that are used to create the `_cta_hashid` so it more faithfully will be created. This was tested locally and the test that was previously failing now passes.


KelseyNewman147 (approved):
LGTM!

PR #365, opened 2024-06-07
clean up mailchimp dbt and migrate to bq v2

emily-flambe:
YOLO-merging this because Mailchimp is just a test sync anyway - we will need new dbt if/when we run this for real.


emily-flambe:
@kanelouise , not sure if this will be relevant to your current work on a new Mailchimp sync, but I'm going ahead and migrating our (very minimal) Mailchimp dbt to be compatible with BQ V2, so that we can try the sync end-to-end using Airbyte Cloud (which requires BQ V2)

PR #369, opened 2024-06-13
feat: make snapchat marketing dbt full-refresh

jitoquinto:
https://techallies.atlassian.net/browse/DE-806

It seems like the merging using the unique keys is doing its job as expected, but moving these models to the full-refresh strat since the airbyte streams are set as full-refresh.

TODO:

- [ ] Remove old partitions from base tables


emily-flambe (approved):
:cat_salute:

PR #370, opened 2024-06-18
hotfix: fb v1 ad_set_report_base should dedupe on hashid

emily-flambe:


huystuhh:
i saw dis live so :shipit:

huystuhh (approved):

PR #375, opened 2024-06-27
echo19: add models for `clicks` table

emily-flambe:


emily-flambe:
YOLO merging because we need to get this QCed in prod today - it's probably fine

PR #376, opened 2024-06-27
hotfix: reconcile conflicting schemas for blocks_incremental

emily-flambe:
There are three tables with fields that need to be removed in order to run this dbt for multiple partners.

registration_forms -> remove field `online_registration`
shifts -> remove `digital`
voter_registration_scan_batches -> remove `scans_need_delivery`

If a client needs those fields in those tables, we will need to rework our sync pipelines to deliver different dbt to different partners (probably by taking this to a private repo).


emily-flambe:
@Frydafly , copying you here for awareness! I'm gonna yolo-merge

PR #378, opened 2024-07-09
feat: take out extra columns from dbt tables

kanelouise:
There were extra columns that were preventing the new additional openfield partners from being great. By yeeting these dbt columns, all tables were able to be created (when running dbt pipes locally).

Once this PR is merged and tested in dev, [this PR](https://github.com/community-tech-alliance/airflow-dags/pull/809/) will be able to be merged into prod!

### DIFF
--- a/dbt-cta/openfield/models/0_ctes/knock_conversation_code_cte1.sql
+++ b/dbt-cta/openfield/models/0_ctes/knock_conversation_code_cte1.sql
@@ -32,8 +32,6 @@ select
     cast(universal_source_code as {{ dbt_utils.type_string() }}) as universal_source_code,
     cast(starting_location_id as {{ dbt_utils.type_bigint() }}) as starting_location_id,
     cast(skip_people_search as bool) as skip_people_search,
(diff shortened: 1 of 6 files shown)


emily-flambe:
@kanelouise, these changes will change the table schemas, right? What happens when we run this dbt on our existing Openfield datasets that have these columns? Will we need to run with `--full-refresh` to update the schemas and then run the DAG with `refresh_matviews: True` to reset the matviews?

Let's also flag this schema change for Partnerships so that they are aware, in case anyone reaches out to ask what happened to the columns. In the event that some partners actually do need this columns, we will probably have to move this dbt to a private repo so that different partners can have different dbt run on their data.

kanelouise:
Thanks for this quick review, @emily-flambe and @jitoquinto. 

Le sigh, seems like this is not the way to fix this, then. I can update the schema in the GCS folder and openfield schema files so that every Openfield sync has these columns, even if they're empty. Better than yeeting the columns and potentially taking away important data for partners.

PR #377, opened 2024-07-03
fix: migrate spoke_mo code to BQ V2

Frydafly:
JIRA ticket with QA passing: https://techallies.atlassian.net/browse/DE-697?atlOrigin=eyJpIjoiYzM0YWMwMzNhYjMwNDFhMDgxYTg5ZjRmMTViYzYyNDYiLCJwIjoiaiJ9 

See two connected PRs: 

- Airflow: https://github.com/community-tech-alliance/airflow-dags/pull/804
- Airbyte: https://github.com/community-tech-alliance/airbyte-configs/pull/154

PR #380, opened 2024-07-11
De 697 migrate spoke spoke mo to bq v2

Frydafly:
this was already approved, I just woopsied and merged too early, reversed that, and now here we go again.

https://github.com/community-tech-alliance/dbt-cta/pull/377


kanelouise (approved):
second time's the charm!

PR #390, opened 2024-08-09
feat: add dbt models for new ab tables

jitoquinto:
https://techallies.atlassian.net/browse/PAD-2426

Add dbt and tests for new AB tables. It doesnt seem like the matview dbt is being used and the cta materialization is only enabled in dev.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,4 +3,4 @@ repos:
   rev: 2.3.0
   hooks:
     - id: sqlfluff-lint
-      additional_dependencies: ['dbt-bigquery==1.6.0', 'sqlfluff-templater-dbt']
+      additional_dependencies: ['dbt-bigquery==1.5.4', 'sqlfluff-templater-dbt']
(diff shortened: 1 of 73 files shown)


chrishenry (commented):
Looks straightforward. what should I be looking for when I review dbt?

emily-flambe:
TLDR: @jitoquinto all the dbt looks good to me, just wanted to flag that the only thing that will actually run in the sync is the new tests added to sources.yml.

@chrishenry some deep lore for you (which is not actionable and probably just confusing): the dbt for Action Builder reflects my vision for delivering data via Analytics Hub while also retaining the option of delivering data as materialized views. dbt doesn't allow you to name different tables the same thing, so the Analytics Hub vs Matview flavors need to have different names. You can see in the Action Builder dbt I opted to add the suffix `_mv` to the materialized views.

But that's actually not even relevant for Action Builder right now, because our DAG doesn't use dbt to deliver data to partners - it only runs the tests in sources.yml.

Why? Our Action Builder DAG originally didn't have any dbt - it just ran some operators to sync data from Postgres => GCS => BQ, and ran a helper function to generate matviews. I added the dbt when I was working on setting things up for an Analytics Hub migration, because the AH strat _does_ require dbt (in the 2_cta_delivery models).

emily-flambe (approved):

kanelouise (approved):
👍

PR #399, opened 2024-08-21
echo19: use existing primary keys instead of hashid for deduping

emily-flambe:
When checking for dupes in echo19 data, I found that the tables all have existing primary keys. We should use these as unique_key instead of hashids.

Also, the tests were missing a few tables, so I added tests for them and included uniqueness tests for the PKs in all tables.

(We don't have any dupes in this sync right now, everything is fine, but we should definitely at least add the tests in.)


huystuhh (approved):
:shipit:

PR #400, opened 2024-08-21
sprout social should use existing primary keys for unique_key (+tests)

emily-flambe:
similar to https://github.com/community-tech-alliance/dbt-cta/pull/399 , which did the same thing for echo19.

Some of the Sprout Social tables have existing primary keys that we should use for the unique_key rather than the hashids. This PR also adds non-null and uniqueness tests for those PKs.


emily-flambe:
> nice! this should also probably update the deduping cte so we dont hit that whole multiple id's from source. I'll do that for the echo19 dbt 🫡

yeah good call I'll update that meaow

PR #402, opened 2024-08-23
fix: remove the videos field from fbpages page table

jitoquinto:
https://techallies.atlassian.net/browse/PAD-2666

FB Pages sync was failing with the following error
```
'{"error":{"message":"An unknown error has occurred.","type":"OAuthException","code":1,"fbtrace_id":""}}'
```
After removing the `videos` field from the page stream, everything seems good


emily-flambe (approved):
yeet

emily-flambe:
@jitoquinto I did already approve this PR but now I'm wondering whether we should just make this _null_ instead of removing? If we remove, the schema will change and we need to drop + rebuild the base table and refresh the matviews. But also that's probably fine lol

PR #405, opened 2024-09-05
feat: add deduping CTEs to snapchat models

emily-flambe:
Snapchat is one of our older syncs, so we hadn't yet gotten around to adding these CTEs to dedupe when there are dupe rows in the raw data (which recently has been the case for `segments`).

Ran this locally, seems fine ✅


chrishenry (approved):
lgtm, thanks for responding so quickly!

PR #410, opened 2024-09-16
clean: add shell script to sqlfluff everything, and then run it

emily-flambe:
This PR has a BIG FLUFFIFICATION of the dbt projects in this repo. To achieve this, I threw together [sqlfluff_all.sh](https://github.com/community-tech-alliance/dbt-cta/blob/ec/fluff/utils/sqlfluff_all.sh), which iteratively fluffs all of the dbt projects. That's really the only file that needs to be reviewed in this PR, because all of the other changes are sqlfluff fixes from running that script. (Good luck trying to review all of the changed files here in Github 🫠 )

It would be cool to run sqlfluff as part of CICD. This script added in this PR isn't appropriate for that because it will always loop through _all_ dbt projects, which takes a while. But this could be adapted for CICD so that it only runs on new/modified files 🧐


kanelouise (approved):
oooo a very very fancy fluffifying. my computer is sputtering while simply trying to hold these 3k+ files. the script looks good! do you want to make a ticket for further CICD science so we don't lose that good idea?

emily-flambe:
> oooo a very very fancy fluffifying. my computer is sputtering while simply trying to hold these 3k+ files. the script looks good! do you want to make a ticket for further CICD science so we don't lose that good idea?

maybe - I've been hesitant to make an actual ticket for linting cicd in this repo because it feels luxuriously low priority, but I mean why not lol

chrishenry (approved):
fluff it up

PR #414, opened 2024-09-26
fix: use our loaded_at column instead of timestamp which is a custom field

jitoquinto:
https://techallies.atlassian.net/browse/PAD-2699

`timestamp` is a kycave custom field. So I added a `loaded_at` column while loading into the raw table that we can use for deduping.


emily-flambe (approved):
ope

PR #433, opened 2024-09-30
feat: new Hustle models for `inbox_*` tables

emily-flambe:
See https://techallies.atlassian.net/browse/PAD-2738

Seems like Hustle added some new tables 🎉 this change adds models to the `hustle` dbt project to normalize and deliver those tables 🎉 yay 🎉 

## TODO before merging:
- [x] merge DAG update to include new tables in the hustle syncs (or else the dbt run step will throw errors I think)


dbt-cta/hustle/models/0_ctes/inbox_addresses_cte1.sql
```diff
@@ -0,0 +1,22 @@
+-- CTE:
```
chrishenry, line 1: what does CTE mean?

chrishenry (commented):
generally lgtm, but I think someone else should prob make the final approval, as I'm still pretty unfamiliar w dbt

huystuhh (approved):
Overall LGTM! Some of the SQL files have an empty line after `select` which will (or well, _should_) fail the sqlfluff lint check now that it's fully operational if you update this PR with changes from the main branch.

dbt-cta/hustle/models/0_ctes/inbox_addresses_cte1.sql
emily-flambe: common table expression. It's just a ~pretentious~ technically accurate way to describe a query that happens before another query but it's all in the same query but you break it up so that the steps are easier for human brains to contemplate 🧠

PR #438, opened 2024-10-07
feat: containerize dbt and add script to build and deploy image

jitoquinto:
https://techallies.atlassian.net/browse/PAD-2963

This PR does the following:
- Add a Dockerfile to containerize dbt
- Adds default values to all the places we are getting Environment Variables in our dbt. This is needed for installing dbt deps when building the Docker image.
- Adds a Makefile that can be use to build image and deploy to Artifact Registry
  - It is using cloud build to build and deploy. This also takes care of the whole having to authenticate to both GHCR and Google for this image.
- Removes gitpod stuff (i dont think anyone is using this anymore, but let me know if i should keep it in there)

Left TODO 

- [x] Add info in README about how to use Makefile
- [x] Figure out if all SQL Fluff errors can be fixed\
  - It looks like some of these errors are because the sql fluff templater were using isnt resolving all the templates we have? So the alias errors that are coming up is because the alias is used in a template. And some other ones are cause some of the macros being used in templates arent being recognized.


README.md
```diff
@@ -58,6 +58,38 @@ If this is your first time running the helper script, make sure to run the `Init
...
 
+## Containerizing dbt-cta
+There is a _Dockerfile_ at the root of this repo that can be used to create a dbt-cta image. This image can be used to run dbt as a container, which makes it possible to run dbt in either Docker or Kubernetes. There is a Makefile that can be used to simplify the process of building and running the dbt-cta image locally for testing!
+
+### Makefile Requirements
+- A Google Cloud Platform(GCP) Project with Cloud Build and [Artifact Registry](https://cloud.google.com/artifact-registry/docs/docker) enabled. The Makefile works by using GCP Cloud Build to build and deploy the image to Google's Artifact Registry. 
+- Setup Docker Authentication to Artifact Registry
+    - [Documentation from GCP](https://cloud.google.com/artifact-registry/docs/docker/authentication)
+    - I've had no issues using the [gcloud cli credential helper](https://cloud.google.com/artifact-registry/docs/docker/authentication#gcloud-helper)
+- If you are running the dbt locally, you mush have Docker up and running on your computer.
```
chrishenry, line 69: typo
```suggestion
- If you are running the dbt locally, you must have Docker up and running on your computer.
```

README.md
```diff
@@ -58,6 +58,38 @@ If this is your first time running the helper script, make sure to run the `Init
...
+    - NOTE: You must have a `.env` file at the root of this repo for this to work. The `.env` file should have the following Variables set.
+    ```shell
+    CTA_PROJECT_ID=source_project_id
+    CTA_DATASET_ID=source_dataset_id
+    PARTNER_PROJECT_ID=destination_project_id
+    PARTNER_DATASET_ID=destination_dataset_id
+    SYNC_NAME=sync_name  # Example: actblue
+    DBT_TARGET="cta"  # Can be any available target in profiles.yml
+    DBT_SELECT="tag:cta"  # Specify what models should be selected, can specify tags, model names, folders, etc..
+    ```
```
chrishenry, line 89: One thing I've done before is to create an `.env.example` file that's kept up to date with the structure. Values or fake, or strings that provide a clue as to where to get potentially sensitive values from.

chrishenry (approved):
This looks great!

README.md
```diff
@@ -58,6 +58,39 @@ If this is your first time running the helper script, make sure to run the `Init
...
+    - [Documentation from GCP](https://cloud.google.com/artifact-registry/docs/docker/authentication)
+    - I've had no issues using the [gcloud cli credential helper](https://cloud.google.com/artifact-registry/docs/docker/authentication#gcloud-helper)
+- If you are running the dbt locally, you must have Docker up and running on your computer.
+
+### Using the Makefile
+Below are the commands you can run with the Makefile. To view a summary of these on your CLI, run `make help`
+
+- `make help` - prints out a help menu with all the available commands
+- `make build PROJECT_ID=myprojectid` - Starts a Cloud Build run to build and deploy an image. It uses the current Git HEAD hash as the tag.
+- `make tag PROJECT_ID=myprojectid` - Adds the _latest_ tag to Artifact tagged with the current Git Hash.
```
chrishenry, line 76: I think I'm seeing a discrepancy with this documentation and current behaviour. I just pushed a tag, but `latest` wasn't applied. 
screen-shot 2024-12-09 at 2 18 49 PM

chrishenry: Just `make build`. Oh, I see, I would've needed to run make tag to update `latest`?

chrishenry: gotcha, makes sense!

ok, I think this is ready to go!

chrishenry (approved):

PR #439, opened 2024-10-07
feat: adding contact attempt for ddx

huystuhh:
From [PAD-2819](https://techallies.atlassian.net/browse/PAD-2819?atlOrigin=eyJpIjoiYWFjOWM3NDI5YjA5NDVmZDljZjc1MDBhNWZmZTkzMmIiLCJwIjoiaiJ9)

[PAD-2819]: https://techallies.atlassian.net/browse/PAD-2819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt-cta/ddx/README.md
+++ b/dbt-cta/ddx/README.md
@@ -2,10 +2,10 @@
 These dbt models are for transforming and delivering DDX standard subscription stream data. Raw data is read from GCS and placed into a temporary raw table that this dbt will use to merge into the CTA incremental base tables. 
 
 # Currently Implemented Streams
+- contact-attempt
 - survey-responses
 - universe-builder
 - universe-builder-raw
-- (Pending) contact-attempt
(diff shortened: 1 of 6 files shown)


huystuhh:
I copied and edited mainly from referencing `survey_response` but I'm not sure about the `subscription_name` column. Is that something that should be in `contact_attempt` as well? @jitoquinto @KelseyNewman147

huystuhh:
I went ahead and added `subscription_name` since it seems like we use that to identify the data once in BigQuery but lemme know if that's incorrect!

huystuhh:
@jitoquinto it looks like the DAG python code does some kind of check for `subscription_name` and `additional_column` (which uses `subscription_name`) so I might just leave it in to keep it chugging 🤔

PR #448, opened 2024-11-21
BQ V2 / Airbyte Cloud migration: Google Directory

emily-flambe:
This PR does the usual set of find-and-replaces (etc) to modify the dbt for the Google Directory sync for compatibility with Airbyte's BQ V2 connector.


emily-flambe:
@jitoquinto, do you have any idea why the SQLFluff lint check is failing? This is what I see when I run SQLFluff locally (via cta_dbt_helper.sh):

image

PR #472, opened 2024-12-10
feat: add workflow to build and deploy dbt-cta image

jitoquinto:
https://techallies.atlassian.net/browse/PAD-2964

Adds GHA workflow to build and deploy the dbt-cta image. The way it is currently set up is 
- Dev
  - On a PR with the `dev_deploy` label, build and deploy a new image whenever there is an update to the PR (new commits pushed). The image will get tagged with the Git SHA
  - We can then use this image by specifying the Git SHA tag
- Prod
  - Is very similar to dev, on a push to main we will build and deploy a new image with the current Git SHA. This will also take the extra step of adding the `latest` tag to the newly built image.

Let me know what you guys think of this strat and if something different makes more sense!

TODO:

- [ ] Add the needed perms to the github-actions service account through TF (currently i did it manually in dev)
  - change `Storage Object Admin` to `Storage Admin`
  - add `Artifact Registry Repository Administrator` 
  - add `Cloud Build Editor`

### DIFF
--- a/dbt-cta/adp_payroll/models/sources.yml
+++ b/dbt-cta/adp_payroll/models/sources.yml
@@ -8,7 +8,7 @@ sources:
   - name: payroll_base
 
 - name: partner
-  database: "{{ env_var('PARTNER_PROJECT_ID') }}"
-  schema: "{{ env_var('PARTNER_DATASET_ID') }}"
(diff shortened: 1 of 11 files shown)


chrishenry (approved):
this looks great! how do we intend it be triggered?

PR #483, opened 2024-12-16
meh: remove the full-refresh flag from dbt-project

jitoquinto:
Having `+full_refresh: false` set in the dbt_project.yml file means we cant pass in full-refresh through the CLI when running. Since we do need to run full-refresh sometimes, I am removing it from the `1_cta_full_refresh` config.


emily-flambe (approved):
yeet

PR #487, opened 2025-03-14
HOTFIX: add status field to inbox_messages table

KelseyNewman147:
Hustle added a new `status` field to their `inbox_messages` table. This PR is for that.


chrishenry (approved):
seems straightforward to me!

PR #488, opened 2025-04-13
try new workflow template

KelseyNewman147:


.github/workflows/run_sqlfluff.yaml
```diff
@@ -24,11 +24,12 @@ jobs:
         environment: dev
         steps:
             - uses: actions/checkout@v4
-            - name: Get all changed sql files
+            - name: Get changed sql files
               id: changed-sql-files
-              uses: tj-actions/changed-files@v45
+              uses: community-tech-alliance/.github/workflow-templates/get-changed-files.yaml@main
```
chrishenry, line 29: probably fine for testing, but we may want to consider an approach where we tag. it'll ensure there's a stable version the next time we decide to make improvements or changes.

KelseyNewman147: Oh yeah sorry this is very much a WIP lol definitely agree a tag would be better long term. Moving this to draft in the meantime!

KelseyNewman147: OK swapped to using tags 👍

huystuhh (approved):
:shipit:

PR #491, opened 2025-04-23
feat: small schema update from Action Network

huystuhh:


chrishenry (approved):
looks reasonable to me

